PR #50, opened 2026-04-22
Fix race in CheckpointManager save/load collectives

### LINKED ISSUE #49: Race in CheckpointManager save and load collectives
`CheckpointManager.save()` returns on non-rank-0 before rank 0 has committed
`train_state.pt`, `metadata.json`, and the `latest` symlink. Post-save readers
on other ranks race these writes, which is flaky on NFS and Lustre because
attribute caching delays when a rank sees rank-0's changes.

`CheckpointManager.load()` independently calls `train_state_path.exists()` on
every rank. On a distributed filesystem those stat results can briefly
disagree. When they do, some ranks enter the `torch.load` branch and others
skip it, then the following `broadcast_object_list` deadlocks because the
non-participating ranks never post the matching collective.

## Fix

- End `save()` with a `dist.barrier()` so every rank observes rank-0's
  filesystem commits before returning.
- In `load()`, let rank 0's `exists()` decide for the group and broadcast the
  flag. Only rank 0 calls `torch.load`; the existing object broadcast
  propagates the loaded state to everyone else.

## Coverage

`tests/distributed/test_checkpoint.py`:
- `test_save_waits_for_rank0_writes`: slows rank-0's `torch.save` and asserts
  every rank measures that latency on return from `save()`.
- `test_load_does_not_hang_on_divergent_exists`: monkey-patches `Path.exists`
  to return False on non-rank-0; `load()` completes instead of deadlocking.

mmshad:
## Summary

- End `save()` with a `dist.barrier()` so non-rank-0 cannot return before rank 0 commits `train_state.pt`, `metadata.json`, and the `latest` symlink. Without this, post-save readers race rank-0's writes, which is flaky on NFS and Lustre due to attribute caching.
- Replace the independent `train_state_path.exists()` on every rank in `load()` with a rank-0-authoritative flag broadcast, so no rank can skip the torch.load / broadcast_object_list path while others enter it (which would deadlock).
- Only rank 0 reads `train_state.pt`; the existing object broadcast delivers the loaded state to the rest.

Closes #49

## Test plan

- [x] `uv run pytest tests/distributed/test_checkpoint.py::TestCheckpointSaveBarrier -v` (2+ GPUs): `test_save_waits_for_rank0_writes` and `test_load_does_not_hang_on_divergent_exists` both pass.
- [x] Full distributed suite green on a 2-node allocation.

### DIFF
--- a/kempnerforge/checkpoint/manager.py
+++ b/kempnerforge/checkpoint/manager.py
@@ -130,6 +130,13 @@ class CheckpointManager:
             # Cleanup old checkpoints
             self._cleanup()
 
+        # save() is a collective: non-rank-0 ranks must not return until
+        # rank-0 has committed train_state.pt, metadata.json, and the
+        # latest symlink. Without this barrier, post-save hooks or readers
+        # on other ranks race rank-0's writes (especially on NFS/Lustre).
+        if dist.is_initialized():
+            dist.barrier()
+
     def wait(self) -> None:
         """Block until any pending async checkpoint save completes."""
         self._async_ckpt.wait()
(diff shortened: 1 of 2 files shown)


Naeemkh (approved):

kempnerforge/checkpoint/manager.py
```diff
@@ -130,6 +130,13 @@ def save(
             # Cleanup old checkpoints
             self._cleanup()
 
+        # save() is a collective: non-rank-0 ranks must not return until
+        # rank-0 has committed train_state.pt, metadata.json, and the
+        # latest symlink. Without this barrier, post-save hooks or readers
+        # on other ranks race rank-0's writes (especially on NFS/Lustre).
+        if dist.is_initialized():
```
Naeemkh, line 137: `dist.barrier()` only make sense when the world size is more than 1. Otherwise, it is extra overhead. It would be good to check world size, too. Not a bug, but reduces the overhead. 

```
if dist.is_initialized() and dist.get_world_size() > 1:
```

Naeemkh (approved):
Looks good to me, please see my comment.

kempnerforge/checkpoint/manager.py
mmshad: Thanks. I think it's moot here: `init_distributed()` in [`distributed/setup.py:155`](https://github.com/KempnerInstitute/KempnerForge/blob/b5dde508929cdd1e78de6fba23bf58b16b14f6cb/kempnerforge/distributed/setup.py#L111) early-returns for `world_size == 1` before calling `init_process_group()`, so `dist.is_initialized()` being True already 
  implies `world_size > 1` throughout the repo.                                                                                                                                                               
                                                                                                                                                                                                              
  Adding the guard to this one line would also make it inconsistent with the sibling `dist.is_initialized()` checks further down in `load()` (the `exists_flag` broadcast and the `train_state` broadcast),   
  and with every other collective guard in the repo (`resilience/health.py:88`, `metrics/tracker.py:83`, `distributed/setup.py:272`), all of which use plain `dist.is_initialized()`.                         
                                                                                                                                                                                                              
  A 1-rank barrier is a dispatch-only no-op (microseconds, once per save), so the gain is not observable. If we want the defensive pattern, better to apply it across all sites in one pass.

PR #52, opened 2026-04-22
Persist dataloader state across checkpoints

### LINKED ISSUE #51: Dataloader state not persisted across checkpoints
`scripts/train.py` passes `scheduler` to `CheckpointManager.save()` but never
passes `dataloader`. Every resume replays the first epoch from scratch:
`StatefulDataLoader.batches_yielded` resets to zero, `MixtureSampler`
re-weighting state is lost, and `DistributedSampler` per-epoch shuffle
position is lost. Repeated or skipped examples result.

## Fix

- Pass `dataloader=dataloader` at both `ckpt_mgr.save(...)` call sites in
  `scripts/train.py` (periodic and shutdown paths).
- The training loop constructs the dataloader after `ckpt_mgr.load()`
  because the loader's identity depends on phase and annealing state that
  `load()` restores. To bridge this, `load()` now stashes dataloader state
  into `_pending_dataloader_state`, and a new
  `CheckpointManager.apply_dataloader_state(loader)` flushes it into the
  freshly built loader.
- Non-stateful loaders (plain `torch.utils.data.DataLoader` for HF
  streaming) are a safe no-op: if `load_state_dict` is absent, the stashed
  state is dropped.

## Coverage

`tests/unit/test_checkpoint.py`: 9 new tests exercising two-phase
save/restore including a full round-trip.

mmshad:
## Summary

- Pass `dataloader=dataloader` to both `ckpt_mgr.save(...)` calls in `scripts/train.py` so `StatefulDataLoader.batches_yielded`, `MixtureSampler` RNG, and `DistributedSampler` epoch position survive resumes.
- Add `CheckpointManager.apply_dataloader_state(loader)` for two-phase restore: `load()` stashes dataloader state, the training loop flushes it after constructing the loader (whose identity depends on phase/annealing state restored by `load()`).
- Non-stateful loaders (plain `torch.utils.data.DataLoader` for HF streaming) are a safe no-op.

Closes #51

## Test plan

- [x] `uv run pytest tests/unit/test_checkpoint.py -v`: 9 new tests including a full round-trip.
- [x] Full unit suite clean.

### DIFF
--- a/kempnerforge/checkpoint/manager.py
+++ b/kempnerforge/checkpoint/manager.py
@@ -60,6 +60,10 @@ class CheckpointManager:
         self._async_ckpt = AsyncCheckpointer(mode=config.async_mode)
         self._process_group = process_group
         self._pp_rank = pp_rank
+        # Dataloader state stashed during load() when the caller cannot yet
+        # provide a dataloader object. Applied later via
+        # apply_dataloader_state() once the loader is constructed.
+        self._pending_dataloader_state: dict[str, Any] | None = None
 
     def _checkpoint_dir(self, step: int) -> Path:
         return self.base_dir / f"step_{step}"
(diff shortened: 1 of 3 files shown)


Naeemkh (approved):
LGTM

PR #56, opened 2026-04-22
Release partial mmaps when MemoryMappedDataset init fails

### LINKED ISSUE #55: MemoryMappedDataset leaks partial mmaps when init fails
`MemoryMappedDataset.__init__` opens one `np.memmap`/`np.load` per file in a
loop and appends to `self._mmaps`. If any open raises partway through, the
prior opens stay reachable through the exception traceback (pytest frames,
`logger.exception`, post-mortem debuggers). Their virtual-memory mappings
remain pinned for the lifetime of the traceback, which in long-lived pytest
sessions or retry loops on Lustre/NFS accumulates until the process runs out
of address space.

The dataset has no `close()` method and no `__del__`, so there is no path to
release the mmaps explicitly either.

## Fix

- Wrap the open loop in `try/except` that calls `_close_mmaps()` on failure
  and re-raises. Partial state no longer leaks through the traceback.
- Add `close()` for explicit release and `__del__` for GC fallback.
- `_close_mmaps` iterates the captured `_mmap` inners and uses
  `contextlib.suppress(BufferError, ValueError)` because:
  - `BufferError` fires when live views into the mapping still exist; drop
    the ref and let GC finish it.
  - `ValueError` fires when another code path already closed the mmap.

## Coverage

`tests/unit/test_dataset_mmap_cleanup.py`:
- `test_partial_open_failure_closes_prior_mmaps`: patches `np.load` to raise
  on the third file, asserts the first two mmaps are closed.
- `test_partial_open_failure_on_bin_files`: same guarantee for the `.bin`
  branch via `np.memmap`.
- `test_close_is_idempotent_and_releases_mmaps`: explicit close + second
  close is a no-op.
- `test_successful_init_keeps_mmaps_open`: regression guard for the success
  path.

mmshad:
## Summary

- Wrap `MemoryMappedDataset.__init__`'s file-open loop in `try/except` that calls `_close_mmaps()` on failure and re-raises, so partial state does not leak through the exception traceback (pytest frames, `logger.exception`, post-mortem debuggers).
- Add `close()` for explicit release and `__del__` for GC fallback.
- `_close_mmaps` suppresses `BufferError` (live views into the mapping) and `ValueError` (already closed by another path) so cleanup is always idempotent.

Closes #55

## Test plan

- [x] `uv run pytest tests/unit/test_dataset_mmap_cleanup.py -v` (4 tests).
- [x] Full unit suite clean.

### DIFF
--- a/kempnerforge/data/dataset.py
+++ b/kempnerforge/data/dataset.py
@@ -198,10 +198,11 @@ class MemoryMappedDataset(Dataset):
         self._mmaps.clear()
 
     def close(self) -> None:
-        """Explicit cleanup for callers that want to release mmaps before GC."""
+        """Release the underlying mmaps. Preferred path; do not rely on ``__del__``."""
         self._close_mmaps()
 
     def __del__(self) -> None:
+        """GC safety net only. Prefer explicit :meth:`close`."""
         with contextlib.suppress(Exception):
             self._close_mmaps()
 
--- a/kempnerforge/data/dataset.py
+++ b/kempnerforge/data/dataset.py
@@ -13,6 +13,7 @@ resumption after checkpoint loads.
 from __future__ import annotations
 
 import bisect
+import contextlib
 import logging
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


Naeemkh (approved):
Relying on __del__ for cleanup is generally discouraged (interpreter shutdown ordering, reference cycles, etc.). The `contextlib.suppress(Exception)` is the right defensive move, but consider documenting that `close()` is the preferred path and `__del__` is a safety net only.

mmshad:
> Relying on **del** for cleanup is generally discouraged (interpreter shutdown ordering, reference cycles, etc.). The `contextlib.suppress(Exception)` is the right defensive move, but consider documenting that `close()` is the preferred path and `__del__` is a safety net only.

Good catch! Pushing a follow-up on this branch: docstring on `__del__` flagging it as a GC safety net, and a tighter close() docstring naming the explicit path as preferred. Happy to also add `__enter__`/`__exit__` for a context-manager path if you want it. Let me know.

PR #60, opened 2026-04-22
Seed all four RNGs on cold start to match checkpoint coverage

### LINKED ISSUE #59: _set_seed covers fewer RNGs than the checkpoint path captures
`kempnerforge/distributed/setup.py::_set_seed` calls `torch.manual_seed`
and `torch.cuda.manual_seed` on cold start. Python's `random` and NumPy's
legacy global RNG are left un-seeded, so they pick up process-level entropy
(PYTHONHASHSEED, wall clock) instead of the configured `seed`.

`torch.cuda.manual_seed` seeds only the *current* device, so on a rank with
multiple visible devices the additional devices also fall back to
process-level entropy.

`kempnerforge.checkpoint.state.get_rng_state` captures all four generators
(python, numpy, torch_cpu, torch_cuda) and `set_rng_state` restores them.
Warm resumes therefore have strictly stronger reproducibility than cold
starts. Any code path using `random.random()` or `np.random.rand()` on
rank > 0 produces different values on cold start than on warm resume of
the same step.

## Fix

In `_set_seed(seed, rank, pp_rank=0)`:

- Call `torch.manual_seed(effective_seed)` (unchanged).
- Replace `torch.cuda.manual_seed` with `torch.cuda.manual_seed_all` so
  every visible CUDA device is seeded.
- Add `random.seed(effective_seed)` (Python stdlib).
- Add `numpy.random.seed(effective_seed)` (NumPy legacy global RNG, the
  same generator `checkpoint.state` captures).

Cold start and warm resume now seed the same set of generators.

## Drive-by

`tests/unit/test_eval.py::TestRunEval::test_perfect_model_low_loss`
depended on `sum(embed(0))` landing positive from inherited global RNG
state. Zero the embedding and set row 0 to ones so the assertion is
deterministic regardless of test ordering.

## Coverage

`tests/unit/test_distributed_seed.py` adds four tests:
- `test_set_seed_seeds_all_four_generators`
- `test_set_seed_varies_with_pp_rank`
- `test_set_seed_same_across_dp_ranks`
- `test_set_seed_matches_checkpoint_coverage`

mmshad:
## Summary

- In `_set_seed`, add `random.seed(effective_seed)` and `numpy.random.seed(effective_seed)` so Python's `random` and NumPy's legacy global RNG are deterministic on cold start. The checkpoint path already captures and restores both, so warm resume is no longer stricter than cold start.
- Swap `torch.cuda.manual_seed` for `torch.cuda.manual_seed_all` to cover every visible device, not just the current one.

Drive-by: `tests/unit/test_eval.py::TestRunEval::test_perfect_model_low_loss` relied on `sum(embed(0))` landing positive from inherited global RNG state. Zero the embedding and set row 0 to ones so the assertion is deterministic regardless of test ordering.

Closes #59

## Test plan

- [x] `uv run pytest tests/unit/test_distributed_seed.py tests/unit/test_eval.py -v`.
- [x] Full unit suite clean.

### DIFF
--- a/kempnerforge/distributed/setup.py
+++ b/kempnerforge/distributed/setup.py
@@ -8,10 +8,16 @@ from __future__ import annotations
 
 import logging
 import os
+
+# ``random`` is aliased to ``_random`` because ``init_distributed`` has a
+# function-local ``import random`` for SLURM-port derivation (a fresh
+# ``random.Random(int(job_id))`` factory isolated from the global RNG that
+# ``_set_seed`` mutates). The underscore keeps the two ``random`` bindings
+# unambiguous when grepping the file.
 import random as _random
 from datetime import timedelta
 
-import numpy as _np
+import numpy as np
 import torch
 import torch.distributed as dist
 from torch.distributed.device_mesh import DeviceMesh, init_device_mesh
@@ -100,7 +106,7 @@ def _set_seed(seed: int, rank: int, pp_rank: int = 0) -> None:
     if torch.cuda.is_available():
         torch.cuda.manual_seed_all(effective_seed)
     _random.seed(effective_seed)
-    _np.random.seed(effective_seed)
+    np.random.seed(effective_seed)
 
 
 def init_distributed(config: DistributedConfig, seed: int = 42) -> DeviceMesh | None:
--- a/kempnerforge/distributed/setup.py
+++ b/kempnerforge/distributed/setup.py
@@ -8,8 +8,10 @@ from __future__ import annotations
 
 import logging
 import os
+import random as _random
 from datetime import timedelta
 
+import numpy as _np
 import torch
 import torch.distributed as dist
 from torch.distributed.device_mesh import DeviceMesh, init_device_mesh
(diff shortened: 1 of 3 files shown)


kempnerforge/distributed/setup.py
```diff
@@ -8,8 +8,10 @@
 
 import logging
 import os
+import random as _random
```
Naeemkh, line 17: Why not just `import random` ?

kempnerforge/distributed/setup.py
```diff
@@ -8,8 +8,10 @@
...
 from datetime import timedelta
 
+import numpy as _np
```
Naeemkh: Same here. 

```python
import numpy as np
```

Naeemkh (changes requested):
Not sure if importing with `_` is a good idea. Please double check.

kempnerforge/distributed/setup.py
mmshad: `_np` was initially for stylistic symmetry but more imports added over time and they don't follow this symmetry anymore. I'll fix it.

kempnerforge/distributed/setup.py
mmshad: The underscore on `_random` is intentional: `init_distributed` has its own local `import random` for the SLURM-port derivation (`random.Random(int(job_id))`), which is deliberately a fresh factory isolated from the global RNG that `_set_seed` mutates. Aliasing the module-level import as `_random` keeps the two random bindings unambiguous when grepping the file.

Naeemkh (approved):

PR #62, opened 2026-04-22
Gate train_state.pt load behind an ownership check

### LINKED ISSUE #61: train_state.pt is loaded with weights_only=False without a trust check
`CheckpointManager.load()` reads `train_state.pt` with
`torch.load(map_location="cpu", weights_only=False)`. The file is a full
pickle (it carries scheduler state, dataloader state, and a
caller-supplied `extra` dict), so any object whose class defines
`__reduce__` runs arbitrary Python at load time.

On shared HPC filesystems, the write side of that file is reachable by
anyone who can place or overwrite a file under the checkpoint directory:

- Group-writable `checkpoint.dir` on a shared lab filesystem: any lab
  member can plant a `train_state.pt`, and the next `latest`-following
  resume executes it in the training process's UID (which holds
  `WANDB_API_KEY`, `HF_TOKEN`, and SLURM credentials via env).
- Checkpoints imported from outside the lab (colleague transfer, HF Hub
  snapshot, tarball from a pretrain run) load with the same trust as
  locally-generated checkpoints.

Same-UID compromise is out of scope (if the attacker can write as you,
they already win), but there is no gate today for the two classes above.

## Fix

Factor the load into a `_load_train_state(path)` helper in
`kempnerforge/checkpoint/manager.py`:

- `stat()` the file and refuse to load if `st_uid != os.getuid()` with a
  `PermissionError`. Imported checkpoints must be `chown`-ed to the
  current user after inspection.
- Warn (but still load) if the file is group- or world-writable, so the
  "someone in my lab can write into `checkpoint.dir`" footgun is at
  least visible in the logs.

This is Tier 1. Tier 2 (convert to `weights_only=True` with an explicit
`safe_globals` allowlist, split scheduler state out of `extra`) is
tracked separately.

## Coverage

`tests/unit/test_checkpoint_security.py`:
- `test_rejects_foreign_owned_file`: crafts a `train_state.pt` with a
  `__reduce__` that would touch a marker file, monkey-patches `getuid` so
  the current UID differs from the file's `st_uid`, asserts
  `PermissionError` is raised *before* the payload fires (marker does not
  appear).
- `test_accepts_own_file`: sanity check that a same-UID benign file still
  loads.
- `test_warns_on_group_writable`: mode `g+w` produces a warning but
  still succeeds.
- `test_no_warning_on_private_mode`: mode `0o600` is quiet.
- `TestManagerLoadRejectsForeignCheckpoint::test_load_raises_before_executing_pickle`:
  full `CheckpointManager.load()` path with `exclude_keys=['model', 'optimizer']`
  (to skip DCP), same outcome.

mmshad:
## Summary

- Factor `torch.load(train_state.pt, weights_only=False)` out into `_load_train_state(path)`. The helper refuses to load when `path.stat().st_uid != os.getuid()` and warns on group- or world-writable files.
- Closes the "group-writable shared checkpoint dir" and "imported pretrained checkpoint" footguns: anyone who can plant a pickled `train_state.pt` under the checkpoint path currently gets code execution in the training process on resume.
- Same-UID attacks remain out of scope (if the attacker writes as you, they already win). This also does not convert to `weights_only=True`; that is Tier 2 (scheduler/extra split + `safe_globals` allowlist), tracked separately.

Closes #61

## Test plan

- [x] `uv run pytest tests/unit/test_checkpoint_security.py -v` (5 tests covering helper + full `CheckpointManager.load()` path).
- [x] Full unit suite clean.
- [x] Sanity resume test (e.g., smoke `TestAutoResume`) still succeeds on a same-UID checkpoint.

### DIFF
--- a/kempnerforge/checkpoint/manager.py
+++ b/kempnerforge/checkpoint/manager.py
@@ -11,7 +11,9 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import shutil
+import stat
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


Naeemkh:
Please address the merge conflict.

Naeemkh (approved):

PR #64, opened 2026-04-22
Enforce per-op timeouts on init and coordination collectives

### LINKED ISSUE #63: Init and coordination collectives ignore declared per-op timeouts
Several collectives outside the training-step reduce path declare or imply a per-op timeout (10s liveness probes, 60s init barriers, 300s eval broadcast, 600s checkpoint broadcast). All of them silently fall through to the process-group default, which is `nccl_timeout_sec` = 1800s. A misconfigured or wedged rank takes 30 minutes to surface during init or coordination instead of the seconds-to-minutes window the call sites expect.

### `kempnerforge/resilience/health.py::check_nccl_health`

The function takes a `timeout_sec` argument and then calls `dist.all_reduce(...)` synchronously. PyTorch ignores the parameter; the all-reduce blocks under the PG default. Every caller asking for a 10s health probe waits 30 minutes when a peer is wedged.

### `kempnerforge/distributed/setup.py` DeviceMesh-construction barrier

The init-time barrier guarding mesh construction is a plain `dist.barrier()` with no per-op bound. A node that fails to join during init holds every other rank for 1800s before NCCL gives up.

### `kempnerforge/training/eval.py` PP eval loss broadcast

The pipeline-parallel eval loss is broadcast from the last stage to every other rank. With no per-op timeout, a stuck eval iteration becomes a 30-minute hang followed by an opaque NCCL watchdog error.

### `kempnerforge/checkpoint/manager.py` train_state broadcast on load

The non-distributed `train_state.pt` is loaded on rank 0 then broadcast via `dist.broadcast_object_list`. A wedged rank or a slow NFS read on rank 0 hangs the whole job for 30 minutes.

### `TORCH_NCCL_ASYNC_ERROR_HANDLING`

When this env var is set to `0` (whether by the user shell or a SLURM prolog), NCCL watchdog timeouts become advisory and may not abort the process group. Per-op timeouts on top of an advisory backend buy nothing.

## Fix

- Switch `check_nccl_health` to `async_op=True` + `Work.wait(timeout=timedelta(seconds=timeout_sec))`. Handle both timeout signaling paths: modern PyTorch raises `RuntimeError`, legacy backends return `False`. Return `False` in both cases so the caller can react.
- Add `_barrier_with_timeout(seconds, reason)` in `distributed/setup.py` and use it for the DeviceMesh barrier (60s). On timeout, raise `RuntimeError` with the reason string attached for diagnosis.
- Wrap the PP eval loss broadcast in a 300s budget and fall back to `nan` loss on timeout so eval reports a clear bad-iteration signal instead of hanging.
- For the train_state broadcast, the original patch wired a 600s `async_op=True` budget. PyTorch 2.11's `dist.broadcast_object_list` does not accept `async_op`, so that change had to be reverted. The other three call sites still get fast-fail; the broadcast inherits the PG default with an explanatory comment, and a wedged rank surfaces via the other paths first.
- In `_set_nccl_env`, `setdefault` `TORCH_NCCL_ASYNC_ERROR_HANDLING` to `1` when unset, and log a loud warning when a user has explicitly set it to `0`. Do not silently override an explicit user choice, but make sure they know it neutralizes every per-op timeout in this patch.

## Coverage

`tests/unit/test_collective_timeouts.py` (10 tests) covers:
- `check_nccl_health` forwards `timedelta` to `Work.wait`, returns `False` on `RuntimeError`, returns `False` on `wait()` returning `False`, returns `True` early when distributed is not initialized.
- `_barrier_with_timeout` raises with the reason string on `RuntimeError`, raises on `False` return, returns normally on success and forwards a `timedelta`.
- `_set_nccl_env` defaults `TORCH_NCCL_ASYNC_ERROR_HANDLING` to `1` when unset, preserves an explicit `1`, warns loudly when set to `0`.

The test file documents that there is no `TestCheckpointBroadcastTimeout` class because of the PyTorch 2.11 `broadcast_object_list` limitation.

mmshad:
Closes #63.

## What

Make declared per-op timeouts on init and coordination collectives actually enforce. Today they fall through to the 1800s process-group default, so a wedged peer hangs the job for 30 minutes regardless of the caller's intent.

- `kempnerforge/resilience/health.py::check_nccl_health` switches to `async_op=True` + `Work.wait(timeout=timedelta(seconds=timeout_sec))` and returns `False` on either timeout signaling path (RuntimeError on modern PyTorch, `False` return on legacy backends).
- `kempnerforge/distributed/setup.py` adds `_barrier_with_timeout(seconds, reason)` and uses it for the DeviceMesh-construction barrier with a 60s budget. Timeouts raise `RuntimeError` with the reason string attached so the failure points at the right phase of init.
- `kempnerforge/training/eval.py` wraps the PP eval loss broadcast in a 300s budget with a `nan` fallback so a wedged PP stage surfaces as a clear bad-iteration signal instead of a half-hour hang.
- `kempnerforge/distributed/setup.py::_set_nccl_env` `setdefault`s `TORCH_NCCL_ASYNC_ERROR_HANDLING=1` and logs a warning when a user has explicitly set it to `0`. Without this, every per-op bound in this patch is advisory.

## Known limit, train_state broadcast

The original commit also wired a 600s `async_op=True` budget around the train_state broadcast in `CheckpointManager.load`. PyTorch 2.11's `dist.broadcast_object_list` does not accept `async_op` (TypeError at runtime), so the second commit on this branch reverts that one site to the plain synchronous form and leaves an explanatory comment. The other three call sites still get fast-fail; a wedged rank surfaces via those before the broadcast hits the 1800s default.

## Tests

`tests/unit/test_collective_timeouts.py` (10 new tests, all passing locally):
- `TestCheckNcclHealthTimeout` covers `Work.wait(timedelta(...))` forwarding, `RuntimeError` and `False`-return timeout paths, and the not-initialized early return.
- `TestBarrierWithTimeout` covers the reason-string error message, the `False`-return path, and the success path.
- `TestNcclAsyncErrorHandlingEnvGuard` covers the unset/explicit-1/explicit-0 cases for `TORCH_NCCL_ASYNC_ERROR_HANDLING`.

The test file documents the absence of a `TestCheckpointBroadcastTimeout` class with a comment pointing at the PyTorch 2.11 limitation.

### DIFF
--- a/kempnerforge/checkpoint/manager.py
+++ b/kempnerforge/checkpoint/manager.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 import json
 import logging
 import shutil
+from datetime import timedelta
 from pathlib import Path
 from typing import Any
 
@@ -28,6 +29,11 @@ logger = logging.getLogger(__name__)
 # Filename for non-distributed training state within a checkpoint directory
 _TRAIN_STATE_FILE = "train_state.pt"
 _METADATA_FILE = "metadata.json"
+
+# Per-operation timeout for the train_state broadcast. Generous enough to
+# cover a cold-NFS torch.load on rank 0, but short enough that a wedged
+# rank surfaces well before the 1800s process-group default.
+_TRAIN_STATE_BROADCAST_TIMEOUT_SEC = 600.0
 
 
 class CheckpointManager:
@@ -183,10 +189,26 @@ class CheckpointManager:
         if train_state_path.exists():
             train_state = torch.load(train_state_path, map_location="cpu", weights_only=False)
 
-            # Broadcast from rank 0 to all ranks
+            # Broadcast from rank 0 to all ranks with a per-op timeout so
+            # a wedged rank surfaces well before the 1800s PG default.
             if dist.is_initialized():
                 object_list = [train_state if self._rank == 0 else None]
-                dist.broadcast_object_list(object_list, src=0)
+                work = dist.broadcast_object_list(object_list, src=0, async_op=True)
+                try:
+                    done = work.wait(timeout=timedelta(seconds=_TRAIN_STATE_BROADCAST_TIMEOUT_SEC))
+                except RuntimeError as e:
+                    raise RuntimeError(
+                        f"train_state broadcast timed out after "
+                        f"{_TRAIN_STATE_BROADCAST_TIMEOUT_SEC}s. If rank 0's "
+                        f"torch.load is slow (cold NFS, large pickle), raise "
+                        f"_TRAIN_STATE_BROADCAST_TIMEOUT_SEC in manager.py; "
+                        f"otherwise a rank is wedged. Underlying: {e}"
+                    ) from e
+                if done is False:
+                    raise RuntimeError(
+                        f"train_state broadcast timed out after "
+                        f"{_TRAIN_STATE_BROADCAST_TIMEOUT_SEC}s."
+                    )
                 train_state = object_list[0]
 
             assert train_state is not None, "train_state broadcast failed"
--- a/kempnerforge/checkpoint/manager.py
+++ b/kempnerforge/checkpoint/manager.py
@@ -12,7 +12,6 @@ from __future__ import annotations
 import json
 import logging
 import shutil
-from datetime import timedelta
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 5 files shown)


Naeemkh (approved):
LGTM

PR #95, opened 2026-05-13
Warn when HF-backed vision encoder is paired with override knobs

### LINKED ISSUE #94: Warn when HF-backed vision encoder is paired with override knobs
HF-backed vision encoders (`siglip2`, `clip`) probe their own
`feature_dim` and `num_tokens` from the loaded HuggingFace model's
config. The corresponding fields in `[vision_encoder]` default to `0`,
which is the "infer at build time" sentinel.

Setting these knobs to non-zero values in the TOML is legal (the
builder applies the user value post-probe), but it silently desyncs
from the actual model. A user who swaps the SigLIP2 backbone for a
different patch size won't realize the old `num_tokens = 196` no
longer matches.

`JobConfig.__post_init__` should warn when:

- `vision_encoder.type in {"siglip2", "clip"}` AND
- `vision_encoder.feature_dim > 0` OR `vision_encoder.num_tokens > 0`.

The `random` stub is excluded because it has no model to probe and the
user-supplied values are the actual values.

Soft-enforce only (warning, not error). The codebase's shipped configs
already follow this pattern: `vlm_7b_siglip2.toml` and
`vlm_7b_siglip2_cross_attn.toml` leave both knobs at `0` and let
SigLIP2-base-patch16-224 probe `feature_dim = 768`, `num_tokens = 196`.

mmshad:
## Summary
- `JobConfig.__post_init__` logs a `WARNING` via
  `logging.getLogger("kempnerforge.config.job")` when
  `vision_encoder.type in {"siglip2", "clip"}` and
  `vision_encoder.feature_dim > 0` or
  `vision_encoder.num_tokens > 0`.
- The HF-backed type set lives at module scope in `config/job.py` as
  `_HF_VISION_ENCODER_TYPES = frozenset({"siglip2", "clip"})`. Adding a
  third HF encoder is one string append.
- The override is still legal (the builder applies the user value
  post-probe). This is a warning, not an error.

Closes #94

## Changes
- `kempnerforge/config/job.py`: `_HF_VISION_ENCODER_TYPES` constant +
  warning in `__post_init__`.
- `tests/unit/test_config.py`: new `TestHfEncoderOverrideWarning`
  class with 5 tests covering siglip2 + num_tokens, clip +
  feature_dim, both at once, HF + zero overrides (silent), random +
  non-zero (silent). Test class flips
  `kempnerforge` logger propagation in `setup_method` because
  `metrics/logger.py:_configure_root` sets `propagate = False` which
  blocks `caplog` from observing records.

## Test plan
- [x] `uv run pytest tests/unit/test_config.py` (111 pass, 1 skipped).
- [x] `uv run pytest tests/unit/` (1232 pass, 2 skipped; was 1227 + 5
      new).
- [x] `uv run ruff check kempnerforge/ tests/ scripts/` (clean).
- [x] `uv run ruff format --check kempnerforge/ tests/ scripts/`
      (clean).
- [x] `uv run pyright kempnerforge/` (0 errors).
- [x] No source-side behavior change for the documented HF path
      (both knobs at 0), so integration/e2e/distributed paths
      unaffected.

### DIFF
--- a/kempnerforge/config/job.py
+++ b/kempnerforge/config/job.py
@@ -18,6 +18,13 @@ from kempnerforge.config.training import TrainConfig
 from kempnerforge.config.vision import VisionEncoderConfig
 from kempnerforge.config.vlm import VLMConfig
 
+# Vision-encoder types whose builders load a HuggingFace model and probe
+# feature_dim / num_tokens from the model's config. Setting these knobs
+# explicitly in the TOML overrides the probe and is almost always a
+# user mistake. The "random" stub is excluded because it has no model
+# to probe and the user-supplied values are the actual values.
+_HF_VISION_ENCODER_TYPES = frozenset({"siglip2", "clip"})
+
 
 @dataclass
 class JobConfig:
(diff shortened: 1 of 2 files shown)


Naeemkh (approved):
Looks good to me.

PR #106, opened 2026-05-18
Rename prep_vlm_coco_smoke.py and add full train/val/test prep

amazloumi:
## Summary
Promote `scripts/prep_vlm_coco_smoke.py` from a smoke-only helper to the canonical COCO-Karpathy prep entry point for VLM training, eval, and smoke runs.

## Testing
- [x] Single-split mode (`--num-samples 8`): bare `Dataset` written; `load_from_disk` round-trip returns correct columns (`image`, `caption`, `references`, `image_id`) and decodes to real PIL.
- [x] `--all-splits` mode (6 samples × 3 splits): `DatasetDict` keyed by `train`/`val`/`test`; captions are real strings in every split; `references` carries 1 entry for train rows and 5 for val/test; `image_id` populated everywhere.
- [x] End-to-end through `HuggingFaceVLMDataset` + `VLMCollator` for all three splits — produces `pixel_values: (B,3,224,224)`, `input_ids/labels: (B,32)`, `image_positions: (B,)`, i.e. the exact contract `scripts/train.py` consumes.
- [x] `tests/unit/test_vlm_dataset.py`: **23 passed**, no regressions.


mmshad:
**doc, non-blocking**: `scripts/prep_vlm_coco.py`:221-224 says the bare-Dataset single-split output matches split="train" by convention. The loader (`kempnerforge/data/vlm_dataset.py`:168) actually ignores split= for the bare-Dataset branch:

```
ds = loaded[split] if not isinstance(loaded, HFDataset) else loaded
```
  
So any value of `data.hf_dataset_split` works equally, the field is unused for single-split outputs. The "by convention" phrasing could mislead readers into thinking "train" is required. Suggested rewrite:

```
# save_to_disk writes a bare Dataset (not a DatasetDict). HuggingFaceVLMDataset
# accepts both via its load_from_disk path; for a bare Dataset it uses the
# loaded object directly, and the data.hf_dataset_split TOML field is ignored.
```

mmshad (approved):

PR #112, opened 2026-05-26
Add configurable data_seed independent of training seed

### LINKED ISSUE #110: Make data-shuffling seed independently configurable from training seed
## What
Allow the data-shuffling seed to be set independently from the parameter-init seed. The sampler classes (DistributedSampler, MixtureSampler) already accept their own seed, but every construction site in scripts/train.py passes tc.seed, so initialization and batch order are always coupled.

## Scope
kempnerforge/config/training.py (new data_seed field on TrainingConfig); scripts/train.py sampler construction sites (L322, L376, L404, and the HF-streaming sampler). Sampler classes in kempnerforge/data/sampler.py already support a separate seed — no change needed there.

## Backward compatibility
Backward compatible. data_seed defaults to None and falls back to seed, so existing configs reproduce their current trajectories exactly.

amazloumi:
## Summary
- Add `TrainConfig.data_seed` (`int | None`, default `None`) so the data-shuffling seed can be set independently of the parameter-init `seed`.
- Add `TrainConfig.effective_data_seed` property — returns `data_seed` when set, else falls back to `seed`, so existing configs reproduce their current trajectory byte-for-byte (willing to remove the backward compatibility).
- Route the five training-data sampler/dataset sites in `scripts/train.py` (VLM, mixture, mmap, HF-streaming, HF-eager) through `effective_data_seed`. Eval samplers keep using `seed`.
- The data seed is passed identically to all data-parallel ranks (unchanged: a consistent global shuffle is partitioned per rank; unlike the init seed in `_set_seed`, it is deliberately not rank-offset).
- Add unit tests for the fallback and override semantics.
  
Enables independently varying parameter initialization and data order, while keeping single-seed runs identical.

## Testing

Closes #110

### DIFF
--- a/kempnerforge/config/training.py
+++ b/kempnerforge/config/training.py
@@ -25,7 +25,8 @@ class TrainConfig:
     max_steps: int = 100000
     grad_accum_steps: int = 1
     grad_clip_norm: float = 1.0
-    seed: int = 42
+    seed: int = 42  # Seeds parameter init, dropout, and (by default) data order
+    data_seed: int | None = None  # Overrides data/batch order only; None falls back to seed
     compile_model: bool = True
     mixed_precision: Literal["bf16", "fp16", "fp32", "fp8"] = "bf16"
     activation_checkpointing: ActivationCheckpointing = ActivationCheckpointing.none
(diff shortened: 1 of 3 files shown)


Naeemkh (commented):
- Looks good to me. I am curious if you can change the seed mid-training. For example, if you train with `data_seed=42` to step 5000, save checkpoints, then edit config to `data_seed=32`, resume. What happens? What should a user expect?
- Are you planning on changing data trajectory mid-run?

Also, I think we need to keep the backward compatibility. The current fallback seems right.

amazloumi:
> * Are you planning on changing data trajectory mid-run?

No; I am curious what would be use cases in that scenario?

amazloumi:
> Also, I think we need to keep the backward compatibility. The current fallback seems right.

Since there is not that much active experiment yet; we may want change all the configs two have both seed and data_seed (even with the same value) then we don't have to have this fallback; but I still fine with keeping the fallback.

Naeemkh:
> > Also, I think we need to keep the backward compatibility. The current fallback seems right.
> 
> Since there is not that much active experiment yet; we may want change all the configs two have both seed and data_seed (even with the same value) then we don't have to have this fallback; but I still fine with keeping the fallback.

Let's do that in a separate PR.

Naeemkh (approved):

PR #113, opened 2026-05-27
Add log-spaced checkpoint schedule

### LINKED ISSUE #111: Add log-spaced checkpoint schedule
## What
Add a log-spaced checkpoint mode (`0, 1, 2, 4, …, log_until`, then every `interval`) for studying early-training dynamics, where dense early checkpoints matter. Fixed-interval saves either explode storage (small interval) or miss the rapid early phase (large interval).

To be useful for a dynamics study, the schedule needs "keep all checkpoints" retention. The runtime already supports this — `CheckpointManager._cleanup()` (`manager.py:618`) returns early when `keep_last_n <= 0` — but `CheckpointConfig.__post_init__` currently rejects `keep_last_n < 1`. So this issue also relaxes that one guard.

## Scope
- `kempnerforge/config/checkpoint.py`: add `schedule: "interval" | "log" = "interval"` and `log_until: int` to `CheckpointConfig`; add their invariants to `__post_init__`; relax the `keep_last_n` guard to allow `<= 0` (meaning "retain all").
- `scripts/train.py:966`: refactor the `step % interval == 0` save decision into a `config.checkpoint.should_save(step)` predicate.
- Docs under `docs/checkpointing/`.
  
## Backward compatibility
Backward compatible. `schedule` defaults to `"interval"`, reproducing current fixed-interval behavior exactly. The `keep_last_n` relaxation only *adds* the previously-forbidden `<= 0` option; existing `>= 1` values are unchanged.

amazloumi:
## Summary
- Add `CheckpointConfig.schedule` (`"interval" | "log"`, default `"interval"`) and `log_until` (default 512). In `log` mode, `should_save()` saves at step 0 and each power of two up to `log_until`, then every `interval` step thereafter — the Pythia/PolyPythias cadence for studying early-training dynamics.
- Add `CheckpointConfig.should_save(step)` and route the save decision in `scripts/train.py` through it (was inline `step % interval == 0`). Default `interval` mode is byte-for-byte unchanged.
- Relax the `keep_last_n` guard to allow `<= 0` = "retain all checkpoints". `CheckpointManager._cleanup()` already early-returns on `keep_last_n <= 0`; this exposes it via config so dense early checkpoints survive retention during a dynamics study.
- Add unit tests for both schedules, keep-all retention, and `log_until` validation.

## Testing

Closes #111

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Configs: `configs/train/vlm_debug_mot.toml` (1-GPU smoke) and `configs/train/vlm_7b_mot.toml` (4-GPU 7B).
 - `install-and-verify` plugin skill: runs `uv sync`, asserts Python ≥ 3.12, then runs the four CI gate checks (`ruff check`, `ruff format --check`, `pyright`, `pytest tests/unit/`). Canonical first command after cloning.
 - `.python-version` pinned to `>=3.12` so uv resolves the interpreter explicitly. Teammates on 3.13 use 3.13 (no download); 3.11-only users get 3.12 auto-fetched.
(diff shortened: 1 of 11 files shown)


Naeemkh (commented):
Please update the CHANGELOG.md file.

Naeemkh (changes requested):
There are three related parameter that is not clear to the user how to use them. 

- In the schedule, without the comment, the user cannot understand if it is base 2 or base 10. May be change the log to `log2`.
- The `log_until`, gives the impression that it is just up to this point, but based on the comment it is only valid for the dynamic logging. 
- If you use `keep_last_n`, regardless of the previous parameters, all logs will be deleted except the last 3 of them. 

Please take a look at them.

Naeemkh (approved):
Looks good to me.

PR #115, opened 2026-05-27
Fix optimizer momentum reset on checkpoint resume

### LINKED ISSUE #114: Resume is not bit-exact: AdamW optimizer state not restored to full precision
## What happened
Resuming from a DCP checkpoint is **state-correct but not bit-exact**. A resumed run does not reproduce the loss trajectory of an otherwise-identical uninterrupted run — even though uninterrupted same-seed runs ARE bit-deterministic (verified: max|Δloss|=0 across 1-GPU / FSDP / EP).

Isolation shows model weights, data-sample position, scheduler, and RNG are restored exactly, but the **AdamW optimizer moments are not**: the first optimizer step after resume diverges, then propagates.

Evidence (tiny dense model, real data, single GPU, bf16, fused AdamW; stop at step 8, resume to 16):
- Pre-resume steps 1–8: interrupted run **bit-identical** to reference → checkpoint-save is side-effect-free; the checkpoint captures the true state.
- Step 9 (first post-resume) `grad_norm` = **0.855 == 0.855** (ref == resumed) → model weights + data restored exactly.
- Step 10 `grad_norm`: ref **0.930** vs resumed **0.973** → the step-9 optimizer update differed → optimizer moments not bit-exact.
- Loss matches steps 9–11, then drifts ≤0.0625 (≈1 bf16 ULP) from step 12.

  ## Reproduction
  Reference (uninterrupted 16 steps) vs interrupt-at-8 + resume-to-16, comparing post-resume losses. Full script in our study worktree: `spikes/resume_test/verify_resume.sh`. Core:
  reference:           --train.max_steps=16 --checkpoint.dir=/tmp/ref  --checkpoint.interval=100
  
  interrupt at 8:      --train.max_steps=8  --checkpoint.dir=/tmp/test --checkpoint.interval=4

  resume to 16:        --train.max_steps=16 --checkpoint.dir=/tmp/test --checkpoint.interval=4

  (debug.toml, dim=128 n_layers=2 vocab=50304, bf16, compile off, seed 7, log_interval=1, real .npy data)

## Observed divergence 
  | step | ref grad_norm | resumed grad_norm | ref loss | resumed loss |
  |---|---|---|---|---|
  | 9  | 0.855 | 0.855 | 10.7500 | 10.7500 |
  | 10 | 0.930 | **0.973** | 10.6875 | 10.6875 |
  | 12 | 1.023 | 0.980 | 10.6250 | **10.5625** |
  | 16 | — | — | 10.5000 | **10.4375** |
  
## Environment
- GPU: NVIDIA H200 (single GPU — reproduces without distributed)
- PyTorch: 2.11.0+cu128
- KempnerForge: commit `2b08f8e`
- bf16 mixed precision, `optimizer.fused=true` (default), AdamW
  
## Likely cause / to investigate in the fix
Optimizer-state round-trip loses precision (AdamW `exp_avg`/`exp_avg_sq`, and/or fp32 master weights under bf16+fused). Check: does `optimizer.fused=false` fix it?
Are fp32 master weights checkpointed? DCP save/load dtype of the optimizer state_dict.
  
## Missing test (add in the fix PR)
No test asserts bit-exact resume — `test_{checkpoint,pp_checkpoint,moe_checkpoint}_resume` only check loss continuity / loss>0, so this passes today. Add a **resume-determinism test** (reference vs interrupt+resume, assert post-resume losses bit-identical) on **single GPU** and **distributed (2-GPU FSDP)**.

amazloumi:
## Summary
- **Bug:** `CheckpointManager` round-tripped optimizer state through raw `optimizer.state_dict()` / `optimizer.load_state_dict()`. On resume the optimizer is freshly constructed, so its `state_dict()` has no `exp_avg` / `exp_avg_sq` tensors yet (AdamW creates per-parameter state lazily on the first `.step()`). `dcp.load` therefore had no moment tensors to fill, the saved moments were silently dropped, and AdamW momentum reset to zero at every resume. Model weights, scheduler, dataloader position, and RNG all restored correctly — only the optimizer moments were lost, so resumed runs were **not bit-exact**.
 DCP's `get_model_state_dict` / `get_optimizer_state_dict` / `set_model_state_dict` / `set_optimizer_state_dict`. The getters build a load template with the moment tensors allocated in the correct FSDP/DTensor layout, so `dcp.load` repopulates them; the setters write them back into the live optimizer. The `exclude_keys` fine-tuning path (load model, skip optimizer) is preserved.
- **Tests** (fail on the pre-fix code, pass after) — single-GPU and distributed, each with a direct moment check and an end-to-end resume check:
- `tests/integration/test_checkpoint_roundtrip.py::TestCheckpointRoundtrip::test_manager_restores_optimizer_moments_single_gpu`
- `tests/distributed/test_checkpoint.py::TestCheckpointRoundTrip::test_resume_restores_optimizer_moments`
- `tests/e2e/test_training_e2e.py::test_resume_determinism_single_gpu`
- `tests/e2e/test_training_e2e.py::test_resume_determinism_2gpu_fsdp`
- The e2e tests use a small **learnable** on-disk dataset on purpose: random tokens give a near-flat loss that masks the momentum reset (the test would pass even on the buggy code).
- **Docs:** added a `### Fixed` entry to `CHANGELOG.md`; corrected `docs/checkpointing/dcp-model.md`, which documented the old raw-`state_dict()` save/load and explained the empty-template behavior as if it were correct.
- **Compatibility note:** optimizer state is now keyed by parameter fully-qualified name rather than positional index. Checkpoints written before this fix will not restore optimizer state on resume (training continues with a fresh optimizer); model state is unaffected. 
    
## Testing
- [x] `uv run pytest tests/unit/ -v --timeout=60` passes (unit suite unaffected)
- [x] `uv run torchrun --nproc_per_node=4 -m pytest tests/distributed/test_checkpoint.py -v` — new moment test passes; existing 8 checkpoint tests still pass
- [x] `uv run pytest tests/e2e/test_training_e2e.py -k resume_determinism --e2e -v` — single-GPU and 2-GPU FSDP resume bit-exact
- [x] Confirmed fail-on-bug: all four new tests fail on `main` (pre-fix), pass on this branch
- Tested on 1 and 2 GPUs (single-GPU + FSDP-2 resume paths)

Closes #114


Naeemkh (approved):
Thanks for fixing it.

PR #123, opened 2026-06-25
Add video understanding to the VLM path

amazloumi:
## Summary
Extends the existing image-VLM path to ingest **video** — a clip is an ordered set of frames — through the same registry-driven, composition-over-inheritance design. Trains from scratch end-to-end on WebVid-10M across **all four fusion archs** (joint-decoder, cross-attention, MoT, MoMa). The text-only and single-image paths are unchanged (bit-exact).

**Pooling connector + token-count plumbing**
- Add `avgpool` and `attentional_pool` (Molmo2-style, mean-query MHA) adapters via `@register_adapter`; introduce a typed `VisionAdapter` base with `output_num_tokens()` so the visual-token count is adapter-derived.
- Thread that count through the build path, the four modality strategies, and the three seq-len checks (`config/job.py`, `distributed/parallel.py`, `model/vlm.py`). Projection adapters stay identity → image path bit-exact.
    
**Video data path**
- `data/video_io.py`: timestamp-based frame sampling (2 fps, uniform, first & last frame kept — Molmo2 §3.1/§A) + PyAV decode. (torchcodec, the paper's decoder, can't load on the cluster — no system FFmpeg + CUDA-lib mismatch — so we use PyAV, whose wheel bundles FFmpeg; lazily imported.)
- `data/video_dataset.py`: `WebVidVideoDataset` (verified `id[:2]/id[:4]/id[:6]/id.mp4` mapping, CSV manifest, reuses the image preprocessing) + `VideoCollator` → `(B, F, 3, H, W)` + frame mask.
- `config/video.py`: `[video]` `VideoConfig` (data_root, split, fps, max_frames, frame_size, max_samples) wired into `JobConfig` (+ `is_video`).
- Adds `av` to dependencies.
- Make the video data side pluggable: [video].dataset_type / sampling_policy select registered builders (webvid / uniform ship), dataset_name de-hardcodes the WebVid corpus dir, and av becomes an optional video dependency group. New dataset styles / policies are additive registrations.
  
**Frame-aware model + training wiring**
- Generalize `_project_image_features` → `_project_visual_features`: folds the frame axis through the encoder + pooler to `(B, F·P′, dim)` (image is the `F == 1` case).
- Thread `frames_per_clip` so the static visual-token count equals `F·P′` (drives the residual budget and MoT's positional split; static == runtime).
- `scripts/train.py` builds the video dataset/collator when `[video]` is set.
- Add `configs/train/vlm_video_webvid.toml` (SigLIP2 + avgpool + WebVid).
  
**Design faithfulness**
- Video is not a new *arch*, so the four existing `ModalityStrategy` classes were generalized rather than duplicated; new components arrive via the registry with no edits to the text/image fast paths.
- FSDP2-only (PP still rejected for VLM); fixed-shape collation for DP-rank consistency; DCP-stable FQNs. 
    
**Deferred to follow-up PRs:**
per-frame timestamp tokens + special tokens, grounding (`<points>`/`<tracks>` + point-F1/track-J&F eval), frame-mask-aware attention, bidirectional visual attention, VLM sequence packing, long-context (blocked on context-parallel being wired), and warm-start from a converted image-VLM checkpoint.
  
## Testing
- [x] `uv run ruff check` passes
- [x] `uv run pytest tests/unit/ -v` — 1493 passed, 2 skipped
- [x] `uv run torchrun --nproc_per_node=4 -m pytest tests/distributed` — 99 passed, 2 skipped, 0 failed (incl. VLM FSDP/MoT/MoMa/cross-attn suites)
- [x] `uv run pytest tests/e2e/ --e2e` — 25 passed, 1 skipped (7B `--slow`). 5 failures (PP / checkpoint-resume / sigterm) are **pre-existing on `main`** — verified identical with this branch's changes stashed; they're in code paths this PR doesn't touch.
- [x] Tested on 4× H100 (FSDP) — from-scratch WebVid training runs end-to-end for all four archs via `configs/train/vlm_video_webvid.toml` (+ per-arch variants).


kempnerforge/model/vlm.py
```diff
@@ -73,20 +73,37 @@ def prepare(
...
-    feats = wrapper.vision_encoder(pixel_values)
-    # Adapter-agnostic dtype lookup: use the first adapter parameter's
-    # dtype so any registered adapter (mlp_2layer, linear, ...) works
+    is_video = pixel_values.dim() == 5
+    if is_video:
+        b, f = pixel_values.shape[0], pixel_values.shape[1]
+        encoder_input = pixel_values.reshape(b * f, *pixel_values.shape[2:])
+    else:
+        encoder_input = pixel_values
+    feats = wrapper.vision_encoder(encoder_input)
```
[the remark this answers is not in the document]
amazloumi: @copilot reevaluate this

kempnerforge/model/adapter.py
```diff
@@ -87,14 +163,148 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        self.o_proj = nn.Linear(in_dim, in_dim, bias=True)
+        self.out_proj = nn.Linear(in_dim, out_dim, bias=True)
+
+    def reset_parameters(self) -> None:
+        for layer in (self.q_proj, self.k_proj, self.v_proj, self.o_proj, self.out_proj):
+            layer.reset_parameters()
+
+    def output_num_tokens(self, num_input_tokens: int) -> int:
+        return pooled_token_count(num_input_tokens, self.pool_window)
+
```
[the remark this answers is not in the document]
amazloumi: @copilot reevaluate this

kempnerforge/config/adapter.py
```diff
@@ -56,14 +60,42 @@ def __post_init__(self) -> None:
...
+        sequence-length checks match the build-time/runtime token budget.
+        Projection adapters are the identity; pooling adapters apply the
+        shared ``pooled_token_count`` math. Non-positive inputs (e.g. the
+        ``num_tokens=0`` "infer at build time" sentinel) pass through.
+        """
+        if num_input_tokens <= 0 or self.type not in self._pooling_types():
+            return num_input_tokens
+        from kempnerforge.model.adapter import pooled_token_count  # noqa: PLC0415
+
+        return pooled_token_count(num_input_tokens, self.pool_window)
```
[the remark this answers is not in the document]
amazloumi: @copilot reevaluate this

kempnerforge/data/video_dataset.py
```diff
@@ -0,0 +1,234 @@
...
+
+import torch
+from torch.utils.data import Dataset
+
+from kempnerforge.data.video_io import decode_video_frames
+from kempnerforge.data.vlm_dataset import (
+    DEFAULT_IMAGE_MEAN,
+    DEFAULT_IMAGE_STD,
+    _pil_to_tensor,
+    _tokenize_and_mask,
```
camilobrownpinilla, line 40: Should we consider promoting `_pil_to_tensor`->`pil_to_tensor` and `_tokenize_and_mask`->`tokenize_and_mask` to since they are used here and in downstream eval ( #124)?

kempnerforge/data/video_dataset.py
```diff
@@ -0,0 +1,234 @@
...
+        self._max_text_len = max_text_len
+        self._max_frames = max_frames
+        self._min_frames = min_frames
+        self._fps = fps
+        self._frame_size = frame_size
+        self._prompt = prompt
+        self._image_mean = image_mean
+        self._image_std = image_std
+        logger.info(
+            "WebVidVideoDataset: %s [%s], %d clips, max_frames=%d, fps=%s, frame_size=%d",
```
camilobrownpinilla: Isn't `webvid-10M` a specific dataset in the WebVid-style? Why are we hardcoding it?

kempnerforge/data/video_dataset.py
```diff
@@ -0,0 +1,234 @@
...
+            frame_size,
+        )
+
+    @staticmethod
+    def _load_manifest(csv_dir: str, max_samples: int) -> tuple[list[str], list[str]]:
+        """Read partition CSVs into (videoid, caption) lists.
+
+        Reads partitions in sorted order, stopping early once ``max_samples``
+        rows are collected so a quick run does not scan the entire 10M-row
+        corpus. ``videoid`` is kept as a string to preserve the digits used by
```
camilobrownpinilla:
```suggestion
        rows are collected so a quick run does not scan the entire
        corpus. ``videoid`` is kept as a string to preserve the digits used by
```
More webvid-10M hardcoding?

kempnerforge/data/video_dataset.py
camilobrownpinilla, line 1: I think this can be generalized a little to support all `WebVid`-style datasets, e.g., by adding a `dataset_name` field in the class init. Right now it seems to be hardcoded for the `WebVid-10M` dataset. 

We may also want to explicitly flag support for more video dataset styles as a follow-up.

kempnerforge/data/video_io.py
camilobrownpinilla, line 1: We should consider generalizing this by adding configurable sampling policies, as I believe this may be a significant design decision for VLMs.

scripts/train.py
```diff
@@ -287,37 +288,69 @@ def main() -> None:
...
+                split=vcfg.split,
+                tokenizer_path=config.data.tokenizer_path,
+                max_text_len=vlm_cfg.max_text_len,
+                max_frames=vcfg.max_frames,
+                min_frames=vcfg.min_frames,
+                fps=vcfg.fps,
+                frame_size=vcfg.frame_size,
+                max_samples=vcfg.max_samples,
+                prompt=vcfg.prompt,
+            )
```
camilobrownpinilla: Another spot to flag for future generalization supposing there exists/we plan to implement different video dataset styles.

camilobrownpinilla (commented):
Overall looks great. The only thing I noticed is that the data side seems to be hardwired to a specific dataset/format (`WebVideo-10M`/`WebVideo`-style); unsure if this was intentional, but perhaps worth flagging.

amazloumi:
> Overall looks great. The only thing I noticed is that the data side seems to be hardwired to a specific dataset/format (`WebVideo-10M`/`WebVideo`-style); unsure if this was intentional, but perhaps worth flagging.

Thanks for the review. Yes, This is a valid point and was intentional for the first working version. My plan to make it pluggable in follow up PRs to avoid long PRs. It is already a large PR, I might consider making it pluggable in the current PR.

kempnerforge/data/video_dataset.py
amazloumi: it can be part of the #124 PR.

kempnerforge/data/video_io.py
```diff
@@ -0,0 +1,116 @@
...
+    pass: each target timestamp is mapped to the first decoded frame at or
+    after it (timestamps past the last frame map to the last frame, so the
+    final frame is always returned). The returned list has length equal to the
+    number of sampled timestamps (``<= max_frames``), or is empty when the file
+    has no decodable video stream.
+
+    Raises whatever ``av`` raises on a missing/corrupt file; callers that train
+    over noisy data should catch and substitute an empty clip.
+    """
+    import av  # lazy: bundled-FFmpeg decoder, optional at import time
```
Naeemkh: If this is optional and lazy import, it should not be part of the dependencies. Please see my other comment.

pyproject.toml
```diff
@@ -15,6 +15,7 @@ dependencies = [
     "datasets",
     "tokenizers",
     "torchao>=0.17.0",
+    "av>=17.1.0",
```
Naeemkh: Let's move this under `[dependency-groups]`, you can add a new `video` section over there.

Naeemkh (changes requested):

kempnerforge/data/video_io.py
amazloumi: Actually I am not sure if we should have it as lazy import. Will move it to specified dependencies if decided to have it optional.

amazloumi: I have moved it yo its dependency group

pyproject.toml
amazloumi: done

Naeemkh (approved):

kempnerforge/data/video_dataset.py
amazloumi: done

kempnerforge/data/video_dataset.py
amazloumi: made it pluggable,

camilobrownpinilla (approved):

PR #124, opened 2026-06-25
Feat/vlm evaluation pipeline

### LINKED ISSUE #122: Add VLM evaluation pipeline
## What
Add support for VLM evaluation by integrating trusted [`lmms-eval`](https://github.com/EvolvingLMMs-Lab/lmms-eval) harness. Direct multimodal analogue to current text-only evaluation using [`lm-eval`](https://github.com/EleutherAI/lm-evaluation-harness). Provides reproducible and comparable numbers across wide variety of benchmarks without building or maintaining our own harness.

There is currently no support for evaluation of VLMs. 

## Scope
Touches:
- `pyproject.toml` — new `[project.entry-points."lmms_eval.models"]` entry (metadata only; **no**
  new runtime dependency — `lmms-eval` is imported lazily).
- `kempnerforge/data/vlm_dataset.py` – promote `_pil_to_tensor` -> public `pil_to_tensor` and extract small `build_tokenizer`     helper.
- `tests/unit/test_vlm_dataset.py` – Change `_pil_to_tensor` calls to `pil_to_tensor`

New:
- `kempnerforge/eval/vlm/` — new subpackage: the `KempnerForgeVLM(lmms)` chat adapter + a
  `ModelManifest`, with loader/preprocessing/generation as helpers.
- `scripts/vlm_eval_harness.py` — new CLI.
- `docs/how-to/run-vlm-evaluation.md` (+ `toctree` wiring).
- `tests/unit/eval/vlm/` (CPU) + a gated 1-GPU integration test.

No new `JobConfig` fields anticipated. Task selection and runtime options are CLI.

## Backward compatibility
Fully backward compatible. New, parallel subsystem.

camilobrownpinilla:
## Summary
Adds a standalone **VLM evaluation pipeline** that evaluates any KempnerForge VLM checkpoint on the standard multimodal benchmarks `lmms-eval` implements as `generate_until` tasks (MMMU, MMBench, ScienceQA, SEED, AI2D, …), by integrating the [lmms-eval](https://github.com/EvolvingLMMs-Lab/lmms-eval) harness through a custom model adapter. The adapter wraps `VLMWrapper` and loads directly from a DCP checkpoint. It registers as an lmms-eval **chat-model plugin** via a `pyproject.toml` entry point, so **lmms-eval stays unmodified** and is **not** added as a dependency. v1 is single-GPU, image-only, and generation-only. **All changes are additive and backward compatible**; the only edit to existing code is a behavior-preserving refactor.

**lmms-eval chat-model adapter** (`kempnerforge/eval/vlm/adapter.py`)
- `KempnerForgeVLM(lmms)`, a chat model (`is_simple = False`) wrapping `VLMWrapper`. Arch-agnostic across the generative arches (joint-decoder / cross-attention / MoT); **MoMa fails fast** — its non-causal expert-choice routing cannot autoregressively generate, and eval requires generation.
- Implements `generate_until` only; `loglikelihood` and `generate_until_multi_round` raise `NotImplementedError` (chat tasks are generation-only — standard multiple-choice benchmarks run as `generate_until`).
- **Cache-less, single-GPU, batched decode.** Re-runs `VLMWrapper.forward` over the growing **right-padded** batch each step — no transformer KV cache (`Transformer.forward` forbids `kv_caches` + any image-conditioning route, and there is no image-conditioned KV-cache path) — reusing `kempnerforge.model.generate.sample`. Requests are grouped by `gen_kwargs` and right-padded to the batch-max length (the layout training uses: image prefix at `0..n-1`, text contiguous from `n`, trailing pads causally masked), so a batched forward gives each row the same real-position logits as decoding it alone (batch-equivalence pinned by a test); EOS / `until` / `max_new_tokens` are tracked per row. Single-GPU is the validated invocation, not baked in: rank/world_size come from the lmms base and model construction sits behind a `_build_model` seam.
- Guards with clear `NotImplementedError` / `ValueError`: MoMa, video/audio, multi-image, multi-turn/few-shot.

**Checkpoint loading + preprocessing reuse**
- Single-process `dcp.load` of the **model shards only** (DCP reshards, so FSDP/PP checkpoints load into the full model); `resolve_resume_path` with a specific-`step_N` fallback; reads plain-JSON `metadata.json` for step/tokens_seen.
- Reuses the exact training-time preprocessing: a behavior-preserving refactor of `kempnerforge/data/vlm_dataset.py` exposes `pil_to_tensor`, `build_tokenizer`, `resolve_pad_id`, and `DEFAULT_IMAGE_MEAN/STD` (the image/text paths stay bit-exact). Prompt rendering flattens the `ChatMessages` text blocks (no chat template, no `<image>` placeholder — images are conditioned via `pixel_values`); model-specific chat-template support is a documented follow-up.

**Registration (lmms-eval stays unmodified)**
- `kempnerforge/eval/vlm/registry.py`: `MANIFEST = ModelManifest(model_id="kempnerforge_vlm", chat_class_path="kempnerforge.eval.vlm.adapter.KempnerForgeVLM")`.
- `pyproject.toml`: a `[project.entry-points."lmms_eval.models"]` entry for auto-discovery — **metadata only**; lmms-eval is not a dependency (install separately with `uv pip install lmms-eval`, mirroring how lm-eval is handled). The eval subsystem is **import-isolated** — `kempnerforge/eval/` is not imported on the default `import kempnerforge` path, so the main package keeps working with lmms-eval absent (pinned by `test_import_isolation.py`); a file-level `# pyright: reportMissingImports=false` keeps `pyright kempnerforge/` green in CI (where the undeclared lmms-eval is absent).

**Harness + docs**
- `scripts/vlm_eval_harness.py`: CLI mirroring `scripts/eval_harness.py` (no conversion). `--config` / `--checkpoint` / `--tasks` required (default suite TBD), plus `--limit` / `--output` / `--device` / `--dtype` / `--batch-size` / `--max-new-tokens`; lazy `lmms_eval.evaluator.simple_evaluate` import with a helpful error if it is not installed.
- `docs/how-to/run-vlm-evaluation.md`, wired into the how-to `toctree`; CHANGELOG.

**Design faithfulness**
- New, parallel subsystem under `kempnerforge/eval/vlm/`; **no changes to training, the existing loss/perplexity eval path, or model code**. Visual input is modeled as an ordered list of frames (a single image is the `F == 1` case) so video is a localized future addition.

**Deferred to follow-up PRs:**
**video**, multi-turn, few-shot, and multi-image tasks **(need the video model / model-side changes)**; `loglikelihood` / `multiple_choice` scoring and multi-round (chat tasks are generation-only); MoMa eval (needs generation support); data-parallel and sharded multi-GPU inference; single image-encode per request (a model-side seam); model-specific chat templates; a representative default benchmark suite; and whether to formalize the lmms-eval dependency.

## Testing
- [x] Ran evaluation end to end on an image-vlm benchmark (single gpu) 
- (N/A) [ ] If distributed code changed: `uv run torchrun --nproc_per_node=4 -m pytest tests/distributed/ -v`
- (N/A) [ ] If training loop / parallelism / optimizers changed: `uv run pytest tests/e2e/ --e2e -v`

Closes #122


kempnerforge/eval/vlm/manifest.py
amazloumi, line 1: Change this to `manifest.py` is become distinguishable from the KF's central registry concept.

pyproject.toml
```diff
@@ -4,41 +4,44 @@ version = "0.1.0"
 description = "PyTorch-native framework for fault-tolerant distributed training of foundation models on AI clusters"
 license = "MIT"
 authors = [
-    { name = "Kempner Institute", email = "kempner_hpc_cluster@harvard.edu" },
+  { name = "Kempner Institute", email = "kempner_hpc_cluster@harvard.edu" },
 ]
 requires-python = ">=3.12"
 dependencies = [
-    "torch>=2.4",
```
amazloumi, line 11: this line and following couple of lines has unnecessary indentation changes. please limit the changes only to the entry point addition.

pyproject.toml
```diff
@@ -4,41 +4,44 @@ version = "0.1.0"
 description = "PyTorch-native framework for fault-tolerant distributed training of foundation models on AI clusters"
 license = "MIT"
 authors = [
-    { name = "Kempner Institute", email = "kempner_hpc_cluster@harvard.edu" },
```
amazloumi, line 7: unnecessary indentation change

pyproject.toml
amazloumi, line 1: throughout the file there are some indentation changes unrelated to the PR. revert those please.

CHANGELOG.md
```diff
@@ -62,6 +62,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Docs: `docs/how-to/run-vlm-evaluation.md`, wired into the how-to `toctree`.
+  - Deferred: single image-encode per request (model-side change), data-parallel and sharded multi-GPU inference, a representative default benchmark suite, whether to formalize the lmms-eval dependency, and confirming frozen vision-encoder weights load from a real checkpoint.
+- **VLM evaluation: batch size > 1.** `KempnerForgeVLM.generate_until` now decodes requests in batches (the `--batch-size` / `batch_size` model-arg) instead of one at a time. Requests are grouped by `gen_kwargs` and the text is **right-padded** to the batch-max length — the same layout training uses (image prefix at `0..n-1`, text contiguous from `n`, trailing pads causally masked) — so a batched forward gives each row the same real-position logits as decoding it alone (pinned by a batch-equivalence test). Each row's next token is read at its own last real position, and EOS / `until` / `max_new_tokens` are tracked per row. Adapter-only; **no model-code changes**. Multi-image / few-shot / multi-turn remain guarded (deferred until the team convenes — they need model-side changes).
```
amazloumi: can we consolidate this with the main change log bullet for this pr.

kempnerforge/eval/vlm/adapter.py
```diff
@@ -0,0 +1,519 @@
...
+)
+from kempnerforge.metrics.logger import get_logger
+from kempnerforge.model.generate import sample
+from kempnerforge.model.vlm import VLMWrapper, build_vlm_wrapper
+from kempnerforge.resilience.elastic import resolve_resume_path
+
+logger = get_logger(__name__)
+
+# Arches whose routing cannot autoregressively generate.
+UNSUPPORTED_GEN_ARCHS = frozenset({"moma"})
```
amazloumi: Lets not excluding one model type hardcoded since it would be against the register-based and pluggable-based of the code base. instead we open an issue for Moma reference feature to add it later.

kempnerforge/eval/vlm/adapter.py
```diff
@@ -0,0 +1,519 @@
...
+    return config
+
+
+def _check_generative_arch(arch: str) -> None:
+    """Fail fast (before building) on arches that cannot autoregressively generate."""
+    if arch in UNSUPPORTED_GEN_ARCHS:
+        raise ValueError(
+            f"VLM arch {arch!r} cannot be evaluated: its routing is non-causal and cannot "
+            f"autoregressively generate, but chat tasks are generation-only. Supported arches: "
+            f"joint_decoder, cross_attention, mot. Generation support for {arch!r} is a tracked "
```
amazloumi: having denylist in eval/adapter and also text listed the supported archs is not maintainable:
(a) so any new non-generative arch is wrongly assumed to support generation and will fail deep in the decode loop instead of failing fast (we need to add them into the list); (b) the "supported arches" text at will silently go stale as arches are added/renamed.

one solution could be: adding a capability flag on the `strategy/VLMConfig` (e.g. supports_autoregressive_generation: bool = True, set False for MoMa) and have the adapter query it, fail-safe (new arches must opt in), and the supported list comes from the registry instead of a literal string.

tests/conftest.py
```diff
@@ -86,6 +89,30 @@ def tiny_job_config(tmp_path) -> JobConfig:
...
 
 
+# ---------------------------------------------------------------------------
+# VLM configs / model
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture
+def tiny_vlm_configs() -> tuple[ModelConfig, VisionEncoderConfig, AdapterConfig, VLMConfig]:
+    """Tiny Joint-Decoder VLM configs for CPU tests (random vision encoder)."""
```
amazloumi, line 100: is this only test JD? if this is the case it would make sense to test it with other models as well.

kempnerforge/eval/vlm/adapter.py
```diff
@@ -0,0 +1,519 @@
...
+            torch.cat([prompts[i], torch.tensor(generated[i], dtype=torch.long, device=device)])
+            for i in range(batch_size)
+        ]
+        real_len = torch.tensor([s.shape[0] for s in seqs], device=device)
+        cur_max = int(real_len.max().item())
+        input_ids = torch.full((batch_size, cur_max), pad_id, dtype=torch.long, device=device)
+        for i, s in enumerate(seqs):
+            input_ids[i, : s.shape[0]] = s
+
+        logits, _ = model(pixel_values, input_ids)
```
amazloumi: The decode loop calls `model(pixel_values, input_ids)` every step, and it re-runs the vision encoder + adapter each time — so the image is re-encoded ~`max_new_tokens` times per request on identical pixels. Combined with the cache-less text recompute, this can dominate runtime on real benchmarks. I see it's noted as deferred, but caching the image embeds per request (encode once, reuse across steps) is far cheaper than the full KV-cache rework and removes the biggest factor — worth considering before this runs at scale.

You can explore the solution for this if it is too complicated it can be moved to follow-up enhancement PRs

kempnerforge/eval/vlm/adapter.py
```diff
@@ -0,0 +1,519 @@
...
+            return -len(args[0]) if isinstance(args[0], str) else 0
+
+        re_ords = Collator(
+            [request.args for request in requests],
+            _collate,
+            group_fn=lambda args: args[2],  # args[2] == gen_kwargs
+            grouping=True,
+        )
+        results: list[str] = []
+        pbar = tqdm(total=len(requests), disable=(self.rank != 0), desc="KempnerForge VLM")
```
amazloumi, line 598: `self.rank != 0` is only used to hide the progress bar here, there's no per-rank request sharding right?

does _load_weights loads a full single-process model?
can you elaborate how the evaluation is done. only one rank is doing all the jobs or all the requests are sharded across ranks and they work together for performance?

amazloumi (commented):
Thanks for this evaluation eval addition. It is a very important part of our workflow. I have made some comments; lets resolve them or adjust the first version and have plan for follow-up PRs.

kempnerforge/eval/vlm/adapter.py
camilobrownpinilla: Yes, `self.rank != 0` is only for gating the progress bar. 

The adapter is single-process and single-GPU. The entire script including `_load_weights` builds a full model in one process and only one rank does all the work.

The `lmms_eval` harness defaults to single-process/single-rank evaluation. They provide support for data-parallel via `Accelerator` or `torch.distributed`, but we currently do not make used of this. I have support for this tracked as a follow-up PR.

kempnerforge/eval/vlm/adapter.py
camilobrownpinilla: Fixed; visual features now cached.

kempnerforge/eval/vlm/adapter.py
```diff
@@ -0,0 +1,628 @@
...
+            torch.cat([prompts[i], torch.tensor(generated[i], dtype=torch.long, device=device)])
+            for i in range(batch_size)
+        ]
+        real_len = torch.tensor([s.shape[0] for s in seqs], device=device)
+        cur_max = int(real_len.max().item())
+        input_ids = torch.full((batch_size, cur_max), pad_id, dtype=torch.long, device=device)
+        for i, s in enumerate(seqs):
+            input_ids[i, : s.shape[0]] = s
+
+        logits, _ = model(pixel_values, input_ids, precomputed_embeds=visual_embeds)
```
amazloumi: we should pass the `frame_mask` here to properly evaluate any clip that decodes to fewer than `frames_per_clip` frames (short videos, or an undecodable clip that zero-pads the whole clip) and not evaluate on fake frames it never saw in training. it could results in wrong benchmark scores. Training in `scripts/train.py` passes `frame_mask`.

kempnerforge/eval/vlm/adapter.py
```diff
@@ -0,0 +1,628 @@
...
+            for args in chunk:
+                # Chat 6-tuple: (context, doc_to_messages, gen_kwargs, doc_id, task, split).
+                doc = self.task_dict[args[4]][args[5]][args[3]]
+                messages = ChatMessages(messages=args[1](doc))
+                frames, prompt = _render_request(messages, self._config.video)
+                if self._is_video:
+                    # Fixed (frames_per_clip, 3, H, W) clip, zero-padded — identical to
+                    # training (frames_to_clip_tensor is the shared helper).
+                    clip, _ = frames_to_clip_tensor(
+                        frames, max_frames=self._frames_per_clip, frame_size=self._frame_size
```
amazloumi: we should the `frame_mask` see my above comment.

kempnerforge/model/vlm.py
```diff
@@ -61,13 +61,19 @@ class ModalityStrategy(Protocol):
...
+    re-running the vision encoder + adapter; when ``None`` (the default) the
+    strategy encodes from ``pixel_values`` as usual.
     """
 
     def prepare(
         self,
         wrapper: VLMWrapper,
         pixel_values: torch.Tensor,
         input_ids: torch.Tensor,
+        precomputed_embeds: torch.Tensor | None = None,
```
amazloumi: Vision projection is arch-independent, but the `precomputed_embeds if not None else _project_visual_features(...)` block is copy-pasted into all four strategies, and an eval-only caching detail now leaks into the general model interface every arch must implement.

We may resolve the cache one level up, in `VLMWrapper.forward`, and have `prepare()` receive already-projected embeds (no pixel_values, no precomputed_embeds, no per-strategy block):
`prepare()` just get the `visual_embeds` instead of (pixel_values and precomputed_embeds) and pass it directly to `ModalityContext(prefix_embeds=visual_embeds,...)`

the `VLMWrapper.forward` also get `visual_embeds`:
```
if visual_embeds is not None:
    assert not self.training, "visual_embeds (cached decode) is inference-only"
    embeds = visual_embeds
else:
    embeds = _project_visual_features(self, pixel_values)
modality = self.strategy.prepare(self, embeds, input_ids, frame_mask=frame_mask)
logits = self.transformer(tokens=input_ids, modality=modality)
return logits, labels
```

amazloumi (changes requested):
The use of `frame_mask` should be added and also the `precomputed_embeds` path that applies to all arch strategies need redesign.

CHANGELOG.md
```diff
@@ -78,6 +78,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
...`tests/unit/eval/vlm/` (CPU) runs against a faithful in-repo fake `lmms_eval` injected via `conftest.py` (`_fake_lmms_eval.py`) — so the adapter/registry tests **always run in CI and contribute coverage** without the undeclared lmms-eval dependency, instead of skipping. `test_adapter.py` covers rendering / preprocessing / `gen_kwargs` / the batched decode loop (`TestGenerateBatchSingle` B=1 regressions + `TestGenerateBatchMulti` batch-equivalence, per-row `max_new_tokens`, per-row EOS) / guards, plus the loader helpers (`_resolve_dtype`, `_load_config`, `_load_weights`, `_log_checkpoint_metadata`, `_first_stop`), the `__init__` guards, and `generate_until` end-to-end; `test_registry.py` covers the manifest. `test_import_isolation.py` still asserts `import kempnerforge` needs no lmms-eval. `tests/integration/test_lmms_eval_contract.py` pins the *real* lmms-eval API and entry-point resolution to the fakes' assumptions (gated on real lmms-eval), and `tests/integration/test_vlm_eval.py` keeps the self-contained DCP round-trip (at `batch_size=2`) + env-gated real-task path. `tests/conftest.py`: `tiny_vlm_configs` / `tiny_vlm_wrapper` fixtures (random vision encoder, CPU, no checkpoint).
```
amazloumi: `test_registry.py` has renamed to `test_manifest.py`

kempnerforge/eval/vlm/adapter.py
```diff
@@ -0,0 +1,697 @@
...
+
+    sampling = temperature > 0
+    top_k = gen_kwargs.get("top_k")
+    top_p = gen_kwargs.get("top_p")
+    return {
+        "until": [u for u in until if u],
+        "max_new_tokens": max_new_tokens,
+        "temperature": temperature,
+        "top_k": (0 if top_k is None else int(top_k)) if sampling else 0,
+        "top_p": (1.0 if top_p is None else float(top_p)) if sampling else 1.0,
```
amazloumi: does this work for `top_p <= 0`? With all tokens masked to `-inf`, the softmax is degenerate and torch.multinomial raises. Nothing catches it (only `_ContextBudgetError` is handled), so the whole chunk's task aborts on a single odd `gen_kwargs` value. Treat a non-positive top_p as disabled.

kempnerforge/eval/vlm/adapter.py
```diff
@@ -0,0 +1,697 @@
...
+                        )
+                        pixels = clip.to(device=self._device, dtype=self._dtype)
+                        mask = fmask.to(device=self._device)
+                    else:
+                        pixels = _frames_to_pixel_values(
+                            frames, self._frame_size, self._device, self._dtype
+                        )
+                        mask = None
+                    prompt_tensor = torch.tensor(token_ids, dtype=torch.long, device=self._device)
+                except (NotImplementedError, OSError, ValueError) as exc:
```
amazloumi: The per-request guard here only catches (`NotImplementedError`, `OSError`, `ValueError`), but on a corrupt or unsupported-codec clip, `_render_request` → `decode_video_frames` raises `av.error.FFmpegError`, whose base class is Exception — not `OSError/ValueError`. So that error slips past this guard and the outer handler at :670 (which only catches `_ContextBudgetError`), goes uncaught, and aborts the entire lmms-eval invocation — losing the scores for every task, instead of skipping the one bad document. Broaden to `except Exception as exc:`. It still logs and skips only the one request (so it's not silent), and it matches the training path.

docs/how-to/run-vlm-evaluation.md
```diff
@@ -0,0 +1,186 @@
...
+
+| Flag | Default | Purpose |
+|------|---------|---------|
+| `--config` | — (required) | KempnerForge TOML the checkpoint was trained with |
+| `--checkpoint` | — (required) | DCP checkpoint dir (run dir or `step_N` dir) |
+| `--tasks` | — (required) | comma-separated lmms-eval task names |
+| `--limit` | `None` | cap examples per task (int count, or `<1.0` fraction) |
+| `--output` | `None` | save full JSON results |
+| `--device` | `cuda` | inference device |
+| `--dtype` | `bfloat16` | model dtype |
```
amazloumi: make sure this default values are consistent with the code.

amazloumi (commented):
Thanks for the changes. It is more robust now. Just some small fixes and it will be good to go.

amazloumi (approved):

PR #172, opened 2026-08-06
Optimize video I/O with seek decoding instead of serial decoding

### LINKED ISSUE #171: Optimize video decode
Given a clip of any length, KempnerForge VLMs sample a small number of frames. However, video I/O currently decodes clips **serially**, decoding an entire clip irrespective of how many frames are to be sampled from it. This unnecessary slowdown makes long video training/eval prohibitively expensive.

Since sampled timestamps are known before decoding, the decoder should seek to the keyframe preceding each target and decode only the short stretch to it, with a serial fallback for containers that can't seek reliably. Frame selection must remain identical to the serial pass so this changes speed only, never model inputs.

camilobrownpinilla:
## Summary

`decode_video_frames` previously decoded every frame of a clip serially to extract $N$ sampled frames. It now seeks to the keyframe at or before each target timestamp and decodes forward only to that frame, so decode cost scales with frames *kept* rather than clip length. Any seek that cannot guarantee identical selection (no `time_base`, missing PTS, seek landing past its target, FFmpeg error) falls back to a single serial pass — the retained reference implementation — so seeking only ever changes speed, never output.

---

### Benchmarks

Frame-selection parity is byte-identical to serial in all cases below. Decode counts via an instrumented container; times single-threaded.

| Clip | serial | seek | speedup |
|---|---|---|---|
| Real WebVid H.264 clip (11.3s, 8 targets) | 282 frames / 128ms | 18 frames / 53ms | **15.7× fewer decodes** |
| Synthetic, 0.4s GOPs (WebVid-like), 8 tgts/12s | 120 | 20 | 6× |
| Synthetic, 3s GOPs (MLVU-like), 8 tgts/30s | 300 | 127 | 2.4× |
| MLVU end-to-end (prior measurement) | — | — | 11–22× |

**Known tradeoff:** dense sampling inside long GOPs re-decodes the GOP prefix per target (e.g. 5s GOPs with 16 targets/30s: 465 vs 300 serial). Real workloads sample sparsely relative to GOP size, so this doesn't occur there; if it shows up in profiling, the fix is a demux-pass keyframe planner, not a heuristic (a stateful-cursor variant was prototyped, benchmarked at zero gain on sparse workloads, and rejected).

## Testing
- [x] `uv run ruff check` passes
- [x] `uv run pyright kempnerforge/` passes
- [x] `uv run pytest tests/unit/ -v` passes (N tests, 0 failures)

Closes #171

### DIFF
--- a/kempnerforge/data/video_io.py
+++ b/kempnerforge/data/video_io.py
@@ -102,13 +102,10 @@ def decode_video_frames(
 
     Reading seeks to the keyframe at or before each target and decodes forward
     (``_decode_seek``), so cost scales with frames kept rather than clip
-    length. Whenever a seek cannot guarantee the same selection — a stream
-    without ``time_base``, frames without PTS, a seek landing past its target,
-    or any FFmpeg error during the attempt — the clip is reopened and decoded
-    in a single serial pass (``_decode_serial``); selection is identical on
-    either path, so seeking only ever changes speed. A clip that is corrupt
-    mid-file consequently raises from the serial retry rather than the first
-    attempt (the same error class surfaces).
+    length. If a seek cannot guarantee the same selection (see
+    ``_decode_seek``), the clip is reopened and decoded in a single serial
+    pass (``_decode_serial``) with identical selection, so seeking only ever
+    changes speed.
 
     Raises whatever ``av`` raises on a missing/corrupt file; callers that train
     over noisy data should catch and substitute an empty clip.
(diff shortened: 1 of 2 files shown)


kempnerforge/data/video_io.py
```diff
@@ -99,29 +119,92 @@ def decode_video_frames(
...
+            return []
         stream = container.streams.video[0]
         stream.thread_type = "AUTO"
         duration_s = _video_duration_seconds(stream, container)
         targets = sample(duration_s, fps, min_frames, max_frames)
+        try:
+            return _decode_seek(container, stream, targets)
+        except (av.FFmpegError, _SeekUnreliableError) as e:
+            reason = f"{type(e).__name__}: {e}"
+    logger.debug("seek decode failed for %s (%s); falling back to serial decode", path, reason)
```
amazloumi: On a corpus that systematically falls back, this is one stdout line per clip per worker, burying the training log. rate-limiting the debug line would be fine.

kempnerforge/data/video_io.py
```diff
@@ -99,29 +119,92 @@ def decode_video_frames(
...
+    with av.open(path) as container:
+        stream = container.streams.video[0]
```
amazloumi, line 154: This is not guarded any more; we can use the same guard as  line 123.
```
if not container.streams.video:
  return []
```

amazloumi (approved):

amazloumi:
Its look good to me. just two minor comment that can be fixed before merging.

Issue #133, opened 2026-06-30
Several tests/e2e/ training tests fail on main (stale checkpoint assertions + PP/resume subprocess failures)


amazloumi:
## What happened
Running the e2e gate (`uv run pytest tests/e2e/ --e2e -v`) on a single 4×H200 node, 6 tests fail. Re-running the failing subset on **clean `main` (3ad9d9a)** reproduces 5 of them, so these are **pre-existing** (surfaced while running the e2e gate for an unrelated PR). Two distinct root causes:

**(A) Stale checkpoint-log assertions — a test bug.** Tests assert the literal log string `"Checkpoint saved"` (and similar), but the training code now logs `"Sync checkpoint saved"` / `"Checkpoint committed"`. The checkpoint *is* written — `test_moe_checkpoint_resume`'s captured output shows `Sync checkpoint saved: .../step_10`, `Checkpoint committed: .../step_10`, and `Training complete: 10 steps` — only the expected substring is outdated. Confirmed for `test_moe_checkpoint_resume`; very likely the same for `test_checkpoint_save_and_resume` and `test_sigterm_triggers_emergency_checkpoint` (the "…was not saved" asserts).

**(B) PP / resume subprocess failures.** `test_pipeline_parallel` and `test_pp_checkpoint_save_and_resume` fail because the launched `train.py` subprocess exits 1 (`assert result.returncode == 0` → `assert 1 == 0`). The child traceback is **not captured** (`error_file: <N/A>`), so the root cause is unknown. A c10d warning is present on this node and may be unrelated: `socket … errno: 97 - Address family not supported by protocol` (IPv6).

**Flaky (not a bug):** `test_hf_dataset_single_gpu` failed once (HF dataset download, returncode 1) but **passed** on the `main` re-run → network-dependent.

## Reproduction
```bash
# clean main (3ad9d9a), single node, 4xH200
uv run pytest tests/e2e/ --e2e -v

# or just the reproducible failures:
uv run pytest \
  tests/e2e/test_training_e2e.py::test_moe_checkpoint_resume \
  tests/e2e/test_training_e2e.py::test_pipeline_parallel \
  --e2e -v
```

## Error / traceback 
```
# (A) stale string -- checkpoint WAS saved:
>   assert "Checkpoint saved" in output, "MoE checkpoint was not saved"
E   AssertionError: MoE checkpoint was not saved
[...]

PR #175, opened 2026-08-25
Add config-declared plugin modules to the config loader

### LINKED ISSUE #174: Config seam so experiments can register their own components
## What

Component registrations only reach the registry because a core module imports them
(`kempnerforge/model/adapter.py`, `kempnerforge/data/video_dataset.py`, ...). An experiment that
ships its own dataset loader, adapter, or loss therefore has nowhere to put it except core — an
`examples/<name>/` experiment would need core to import `examples/`, which breaks the rule that
experiments are standalone and deleting one has zero impact on core.

Importing the experiment's module by hand before the run does not work either: several sub-config
`__post_init__`s resolve names against the registry at construction time (`kempnerforge/config/video.py`
rejects a `dataset_type` not in `registry.list_video_datasets()`; `AdapterConfig` and
`VisionEncoderConfig` do the same), so anything imported after the config is built is too late.

Add a top-level `plugins` list on `JobConfig`:

```toml
plugins = ["my_experiment.components"]

[adapter]
type = "my_connector"   # registered by the module above
```

`load_config` imports each module *before* the TOML dict and CLI overrides are applied to the
dataclasses, so registrations land before the validators run. A missing or failing module raises a
`ValueError` naming the module and the config key, before distributed init, so a typo dies
immediately on every rank rather than after NCCL bring-up.

## Scope

- `kempnerforge/config/job.py` — the `plugins` field and its `__post_init__` validation
- `kempnerforge/config/loader.py` — importing the modules ahead of both overlay layers

No new registry mechanism: registrations still arrive through the existing `@registry.register_*`
decorators, and core never imports `examples/`.

## Backward compatibility

Backward compatible. `plugins` is a new optional field with an empty default: when it is absent no
import machinery runs, and existing configs, checkpoints and launch commands behave identically.
`--plugins=["pkg.mod"]` overrides the TOML list like any other field.

amazloumi:
## Summary
- Add `JobConfig.plugins`: a top-level list of module paths, empty by default.
- `load_config` imports those modules *before* the TOML dict and CLI overrides are applied to the dataclasses, so components an out-of-tree module registers are already in the registry when the `__post_init__` validators resolve names against it (`adapter.type`, `video.dataset_type`, `vision_encoder.type`, ...). Importing after the overlay would be too late.
- This is the seam that lets an experiment under `examples/` ship its own dataset, adapter, or loss without any core module importing it.
- A missing or failing module raises a `ValueError` naming the module, the config key, and the underlying cause — raised in `load_config` (`scripts/train.py:77`), before `init_distributed` (`:83`), so a typo dies immediately on every rank instead of after NCCL bring-up.
- `plugins` is rejected in `__post_init__` unless it is a list of strings; `--plugins=["pkg.mod"]` overrides the TOML list like any other field.
- Docs: the key, the `sys.path` rules, and four rules for plugin authors (rank-invariant registration, no RNG use at import, no blocking I/O or collectives at import, keep `plugins` in the config you resume with) in `docs/configuration/registry.md`, plus a `JobConfig` row in `config-sections.md` and a CHANGELOG entry.
- A plugin cannot capture a core component name. Under the real entry point 12 of the 14 registry categories are already populated when `load_config` runs, so a colliding registration fails at `scripts/train.py:77` with `Failed to import plugin module '<mod>' from config key 'plugins': ValueError: <category>/<key> is already registered` — rank-uniform and before `init_distributed`. The two categories registered lazily in `VideoConfig.__post_init__` (`video_dataset`, `sampling_policy`) collide on core's later import instead: still loud, still pre-NCCL, still on every rank, but the message names core's key rather than the plugin. No silent-miscompute path exists in either case.
- Known limitation, documented: the list is not recorded in the checkpoint, so a resume that builds a different component fails the DCP load with `Missing key in checkpoint state_dict` unless the new component's parameters are a strict subset of the saved ones. Recording it the way `vlm_freeze` metadata already is would close that; left as a follow-up rather than widening this PR into the checkpoint layer.

## Testing
- [x] `uv run ruff format --check kempnerforge/ tests/ scripts/` passes (176 files)
- [x] `uv run pytest tests/unit/ -v` passes (1697 passed, 3 skipped; +15 tests, 1685 -> 1700 collected)
- [x] `uv run pytest tests/integration/ -v` passes (77 passed, 2 skipped)
- [x] `uv run torchrun --nproc_per_node=4 -m pytest tests/distributed/ -v` passes (99 passed, 2 skipped; `test_checkpoint.py` + `test_resilience.py` on their own: 13 passed)
- [x] `uv run pytest tests/e2e/ --e2e -v` — 5 failed, 26 passed, 1 skipped. The same five test IDs fail on a clean `origin/main` checkout in the same allocation (sorted diff of the failure sets is empty), and all five are #133 with both root causes diagnosed there. Three fail on a stale assertion — they grep for the literal `"Checkpoint saved"`, which `kempnerforge/` never emits (it logs `Checkpoint committed`). Two die in the training subprocess on `AttributeError: 'PipelineStageModule' object has no attribute 'cross_attention_layers'`, reproduced on `17248f8`, this PR's base commit. `test_pp_checkpoint_save_and_resume` carries both defects: it never reaches its assertion, so it needs both fixes.
- [x] `uv run pytest tests/smoke/ --smoke -v` — 1 failed, 15 passed, 3 skipped. The single failure `TestTrainFeatures::test_dense_pp2` reproduces on the clean `origin/main` checkout (same test ID) with that same `PipelineStageModule` signature. So the pipeline-parallel defect accounts for three failures across two suites — one known cause (#133), in code this PR does not touch.
- [x] `uv run make -C docs strict` — build succeeded
- [x] Tested on 4 GPUs with config: `plugin_vlm.toml` — a VLM config whose `[adapter] type = "plugin_probe_mlp"` and `train.loss_fn = "plugin_marker_ce"` come from a module that lives outside the repo, selected only by `plugins = ["kf_plugin_demo"]`. 140 steps under FSDP2, loss 10.75 -> 5.66, no NaNs; the log shows the module imported, the plugin adapter built (`in_dim=384 out_dim=256`), its `forward` running on `(2, 64, 384)`, and the plugin loss called.
- [x] Tested on 4 GPUs with config: the same file plus `--plugins=[]` — all 4 ranks fail with `Unknown adapter.type: 'plugin_probe_mlp'. Registered: [...]`, and with `--plugins=["kf_no_such_module"]` all 4 ranks fail with the plugin-import `ValueError` in 14s. Both stages log 0 `DeviceMesh` lines and 0 NCCL lines, i.e. they die before distributed init.
- [x] Tested on 4 GPUs with config: `configs/train/debug.toml --data.hf_dataset_name=wikitext --data.hf_dataset_config=wikitext-2-raw-v1 --data.tokenizer_path=gpt2 --model.vocab_size=50257 --train.max_steps=60` — with no `plugins` key, every logged `loss`/`lr`/`grad_norm` triple is identical to the same run on a clean `origin/main` checkout (`[step 1] loss=10.8750` -> `[step 60] loss=8.2500`). A `JobConfig` repr sweep over 34 shipped configs is likewise unchanged. The seam is inert when unused.
- [x] Preemption drill on the plugin config (4 GPUs): SIGTERM mid-run -> emergency checkpoint written at step 99 (`.metadata`, `metadata.json`, `train_state.pt`, 4 `.distcp` shards, `latest`) -> auto-resume at step 99 -> steps 100-140 identical in loss/lr/grad_norm to an uninterrupted 140-step run of the same config. Resuming that checkpoint with the built-in `mlp_2layer` adapter and no `plugins` fails loudly on all ranks: `Missing key in checkpoint state_dict: model.adapter.proj1.weight`.

Review note: the repository's `code-review` skill ran over this diff; its findings were verified independently and none were blocking.

Closes #174


Naeemkh (approved):
One follow-up that shouldn’t block this work but is worth tracking: the checkpoint doesn’t record plugins. As a result, when resuming, a component with fewer parameters than the saved version may load without reporting an error. Maybe open an issue to double-check this later.

amazloumi:
Filed as #180, with a reproduction.

Worth noting the two shipped adapters do not exhibit it — `linear` uses `proj.*` and `mlp_2layer` uses `proj1.*`/`proj2.*`, so those names are disjoint and a swap fails loudly in both directions, which is what `TestComponentSwapOnResume` pins. Constructing a genuine name-wise subset reproduces exactly what you described: saving `proj1.*` + `proj2.*` and resuming into a component with only `proj1.*` loads silently and restores step 7, dropping `proj2` with no error.

#180 records that, notes the `plugins` seam is what makes overlapping names reachable (a plugin's parameter names are unconstrained, unlike core's reviewed set), and suggests recording the resolved list in `metadata.json` the way `vlm_freeze` already is.

Issue #177, opened 2026-08-25
Emergency checkpoint is lost on SIGTERM when data.num_workers > 0, and auto-resume then fails


amazloumi:
## What happened

On SIGTERM (SLURM preemption), the emergency checkpoint is **silently not written** when `data.num_workers > 0`, and the run is then **unable to auto-resume**.

Two distinct defects compose:

**(1) The shutdown handler never fires.** SIGTERM reaches the DataLoader worker processes as well as the main process. The workers die, and the main process fails fetching the next batch before `ShutdownHandler` can run the emergency save. The log shows no `Shutdown requested at step N — saving emergency checkpoint` line at all, and no checkpoint is written.

**(2) Auto-resume then walks into the wreckage.** `resolve_resume_path` (`kempnerforge/resilience/elastic.py`) applies **no durability check**: it returns the `latest` symlink, or failing that the highest-numbered `step_N` directory, with no test that DCP's `.metadata` is present. `CheckpointManager` only ever points `latest` at a durable checkpoint (#102), but the `step_N` fallback has no equivalent guard. A run whose only checkpoint directory is a failed emergency save therefore resumes into it and dies with `AssertionError: metadata is None`.

With `data.num_workers = 0` the same drill is clean: the handler fires, the checkpoint is written and committed, and resume restores both trajectory and dataloader position.

## Reproduction

Pre-existing on clean `main`; reproduced with a stock config.

```bash
# clean main, 4xH200, single node
uv run torchrun --nproc_per_node=4 scripts/train.py configs/train/vlm_debug.toml
# send SIGTERM mid-run, then relaunch the same command to trigger auto-resume
```

A controlled A/B isolates `num_workers` as the cause — same commit, same node, same drill, one variable:

| leg | `data.num_workers` | rc | `Shutdown requested` lines | outcome |
|---|---|---|---|---|
| dense | 0 | 0 | 1 | emergency save at step 119; resumed at `epoch=7, skip_batches=14`; completed 122 steps |
| vlm | 2 | 1 | 0 | handler never fired; workers killed; save never started |

## Error / traceback

```
# (1) with num_workers > 0 -- no "Shutdown requested" line is ever emitted,
#     and no "Checkpoint committed" / "Sync checkpoint saved" line follows.

# (2) on the next launch, auto-resume selects the empty step_N directory:
AssertionError: metadata is None
```

## Suggested fix

The two halves are independently worth fixing:

[...]

PR #178, opened 2026-08-25
Decompose scripts/train.py:main() into kempnerforge/training

### LINKED ISSUE #130: Decompose scripts/train.py:main() into cohesive, testable units
## What
main() in scripts/train.py is ~1000 lines that mix many responsibilities — config load/validate, distributed setup, model/optimizer/scheduler build, checkpoint & resume, the training loop, teardown, and resilience — in a single scope. Break it into named, single-responsibility units called by a thin main() orchestrator.

## Benefits: 
- Readability / maintainability — each phase is understandable and changeable in isolation; the shared scaffold lives in one place (smaller bug-fix surface).
- Unit-testability — extracting the loop/teardown into run_training_loop(...) (or a small Trainer) creates a seam to inject collaborators (e.g. a spy CheckpointManager), so save/exit decisions can be unit-tested fast and in default CI. Today the loop is only reachable via the opt-in --e2e tier — e.g. the #129 final-checkpoint fix has no per-PR coverage.
- Reuse — clean phases are reusable by other entry points (eval, smoke harnesses) without copy-paste.
  
Suggested units (illustrative): load_and_validate_config, setup_distributed, build_model_and_optim, setup_checkpointing, and run_training_loop(...) holding the loop + teardown, with the PP/VLM/standard model-call as an injected step function.

## Scope
- scripts/train.py — main() becomes a thin orchestrator; phases move to functions.
- Optional new home: kempnerforge/training/loop.py (and siblings).
- tests/unit/ — new unit tests enabled by the seam.
- No config sections change; CLI/UX unchanged.

## Non-goals 
- Do not intend to split into per-model scripts (train_dense.py / train_moe.py / train_vlm.py). That duplicates the shared scaffold and contradicts the registry/config-driven design ("swap arch without touching scripts/train.py"; loop variants differ only at the model-call site). Keep one entry point; model differences stay behind the registry + a thin step seam.

## Backward compatibility 
Behavior-preserving refactor — no functional change. Same CLI (uv run torchrun … scripts/train.py configs/train/<file>.toml …), same single entry point, same config/registry dispatch. All branches must stay equivalent: PP / VLM / standard step; NaN-rollback, NaN-continue, NCCL-health, graceful-shutdown exits; phase scheduling; metrics/profiler/eval; and the save → wait order. tests/e2e/test_training_e2e.py must pass unchanged as the safety net.

## Acceptance criteria 
- [ ] main() reduced to a thin orchestrator; each phase a named unit; pyright + ruff clean.
- [ ] tests/e2e/ --e2e passes unchanged (proves no behavior change).
- [ ] run_training_loop(...) unit-testable fro example with a fake CheckpointManager (no real DCP/subprocess);

### LINKED ISSUE #181: Pin the dp_size and phase-temperature propagation paths introduced by the training-entry extraction
## What

#178 routes two values through new dataclass plumbing that nothing currently pins. Both propagation paths work today — the multi-node run reports `Training complete: 10 steps, 163,840 tokens`, which is only correct if `dp_size` reached the counter — but a future regression in either would be **silent**: training completes, loss is unaffected, and nothing raises.

**1. `dp_size` propagation.** `get_dp_info()` → `DataPipeline.dp_size` → `tokens_in_step = batch_size × seq_len × grad_accum_steps × data.dp_size`. Previously `dp_size` was a local used directly at the call site; the round-trip through the dataclass is new. If a builder stops setting it, the field defaults to `1` and `tokens_seen`, `tokens_per_sec` and MFU are wrong by the DP factor on every multi-GPU run. Every existing test in `tests/unit/test_train_entry.py` constructs `DataPipeline()` at its default, so none can distinguish the two cases.

**2. `PhaseState.temperature` propagation.** `config.data.mix_temperature` → `PhaseState.temperature` → `MixtureSampler.update_weights(..., temperature=)`. Previously the config value was passed directly at each call site; hoisting it onto the dataclass is new. `FakeMixtureSampler` already *records* the temperature it receives and no assertion reads it. The default is `1.0`, so a regression that falls back to the default is invisible under stock configs and silently changes mixture sampling under tuned ones.

Note (2) is a **pinning** gap rather than a coverage gap — those lines are already green, so coverage tooling will never flag it. The line executes; nothing fails if it is wrong.

## Scope

`tests/unit/test_train_entry.py` only. Two focused tests, no production code:

- monkeypatch `get_dp_info` to return a non-unit `dp_size`, then assert the value reaches `tokens_seen` — failing if the plumbing is dropped anywhere along the path
- assert `FakeMixtureSampler`'s recorded temperature equals the configured `mix_temperature` after a phase transition, rather than merely recording it

Both are CPU-only and require no GPU tier.

## Backward compatibility

Not applicable — tests only, no production code and no config surface.

## Not in scope

The five `-O`-safe raises added in #178 (`pp_group`, `build_model`, `pipeline_step`, `_build_hf_pipeline`, `BatchStream`) are new and uncovered, but four guard states that `JobConfig.validate` and `select_step_fn` make unreachable, and all five fail loudly if reverted. Testing them would pin a contract whose violation has no observable consequence.

The remaining uncovered lines in `kempnerforge/training/` are code that moved out of `scripts/train.py`, which `[tool.coverage.run] source = ["kempnerforge"]` never measured, and which the GPU tiers exercise. That is a separate initiative, not this work.

amazloumi:
## Summary
- Move `scripts/train.py:main()` (~1000 lines) into `kempnerforge/training/`: `runtime.py` (`RuntimeContext`, `PipelineBundle`, `setup_distributed`, `pp_group`), `data_pipeline.py` (`DataPipeline`, `PhaseState`, the data/eval/phase builders), `loop.py` (`BatchStream`, `StepResult`, `TrainingSession`, `run_training_loop`, and the `text_step` / `vlm_step` / `pipeline_step` bodies picked once by `select_step_fn`), `entry.py` (`run_training` plus `build_model` / `build_checkpoint_manager` / `restore_checkpoint`).
- `scripts/train.py` is now argv + `load_config` + `run_training(config)` — same CLI, same `main()`, same usage string and exit code.
- `run_training(config, *, step_fn=None, hooks=None)` lets an experiment own the step body or register hooks without copying the build phases; both default to today's behavior.
- Teardown is explicit now that this is a callable API rather than a script about to exit: `tracker.close()` always runs, and the collective teardown (`ckpt_mgr.wait()`, which ends in `dist.barrier()`, and `destroy_process_group()`, which blocks in `_wait_for_pending_works()` and shuts NCCL backends down in order) stays on the success path — running either while an exception unwinds meets peers that are still mid-collective and converts a one-rank failure into an all-rank hang.
- `kempnerforge/checkpoint/` is untouched, so DCP key names, layout and the save -> wait order are byte-identical.
- `freeze_meta_at_step` gives save and resume one expression for the VLM freeze metadata; `_resolve_pad_id` reuses `data/vlm_dataset.resolve_pad_id`.
- New `tests/unit/test_train_entry.py` (51 tests): drives `run_training_loop` with a fake `CheckpointManager` (save schedule, off-schedule final save, NaN rollback, shutdown emergency save, resume accounting, and that the exception path runs no collectives), plus `run_training`'s step_fn/hooks/teardown wiring, phase scheduling, `BatchStream`, step dispatch and the `DataPipeline` invariants.
- Doc pointers updated where the loop moved (training-loop, data-flow, architecture index, training index, hooks, resilience, debug-training-regressions, CHANGELOG, codebase-map).

## Incidental fixes
- **`build_eval_dataloader`'s HF-eval broadcast** is now guarded by `dist.is_initialized()`. Extracting it made it a public builder reachable single-process (its own unit test calls it that way), where the unguarded `dist.broadcast` was a guaranteed `RuntimeError`.
- **`pp > 1` with `[[data.datasets]]`** raised `NameError: ds_loss_sums` at the first logging step; `StepResult`'s empty-dict defaults make it report empty per-dataset metrics instead. **Static claim plus unit coverage, not a GPU-reproduced one.** On `main`, `ds_loss_sums` is bound only inside the non-PP step body (`scripts/train.py:798`) while the per-dataset metrics block reads it unguarded (`:975`). I could not reproduce it on GPU because every `pp > 1` run dies earlier at build (see below), so it stays unverifiable there until that is fixed.

## Behavior parity
All numbers below come from one SLURM allocation that self-certified `git rev-parse HEAD == 8aaa72b` with empty `git status --porcelain` for both this branch and a clean `origin/main` worktree, and aborts rather than emitting output otherwise.

- **Loss curves, 4xH200, 20 steps, `--metrics.log_interval=1`, comparing the `(loss, lr, grad_norm)` triple at every step.** `main` was run **twice** first, to establish that bit-exact is the achievable bar on this hardware — without that noise floor, "main vs branch is identical" could just mean the config has low variance. All nine comparisons are bit-exact: three noise floors (main run1 vs run2) and three main-vs-branch, across `debug.toml` (10.4375 -> 10.4375), `debug_moe.toml` (10.4578 -> 10.4579) and `vlm_debug.toml` (10.9375 -> 8.3125).
- **Final checkpoint weights, `main` vs branch, tensor by tensor via `dcp_to_torch_save` + `torch.equal`:** dense 156/156 bit-equal, MoE 236/236, VLM 172/172.
- **Resume continuity** on a real on-disk dataset: interrupt at step 10, resume to 20 — steps 11-20 identical to the uninterrupted run, final weights 156/156 bit-equal.
- **Cross-version resume, both directions:** a checkpoint written by `main` and loaded by this branch, and one written by this branch and loaded by `main`, each reproduce the `main`-only reference over steps 11-20 with 156/156 bit-equal final weights. **The reverse leg is the DCP-contract proof** — a renamed or reordered checkpoint key would break `branch-written -> main loads`, which forward-only resume would not catch.
- **Preemption (dense):** SIGTERM mid-run -> `Shutdown requested at step 119`, `Checkpoint committed: .../step_119`, `Graceful shutdown complete`, exit 0; auto-resume restores `epoch=7, skip_batches=14` and continues from step 119. The emergency `train_state.pt` carries `dataloader={'epoch': 7, 'batches_yielded': 14, 'sampler': {...}}`, so the loader resumes at the position it saved rather than at an epoch boundary. I ran the same drill on `vlm_debug.toml`, where the handler never fired (`Shutdown requested` count 0 vs 1 for dense). The two legs are a controlled A/B on one variable — same branch, same commit, same node, same drill, differing only in `data.num_workers` (dense overrode it to 0; `vlm_debug.toml` leaves it at 2) — and they are the causation evidence in #177.
- **Static:** AST diff of the old `main()` against the new modules — all 29 `logger.*` / `raise` message templates identical, and every collective-bearing call at the same count under the same rank-uniform guard.

## Pre-existing failures, verified on `main` in the same allocation
`tests/e2e/` and `tests/smoke/` have failures that reproduce identically on clean `origin/main`. Two distinct causes, and the second is a root cause #133 could not previously obtain:

1. **Stale assertions** (#133, category A). Four e2e tests assert the literal `"Checkpoint saved"` (`tests/e2e/test_training_e2e.py:387,530,606,716`). That string does not exist anywhere in `kempnerforge/`. The mode-independent replacement is `"Checkpoint committed"` (`checkpoint/manager.py:195`); `"Sync checkpoint saved"` only fires under `async_mode = disabled` (`checkpoint/async_save.py:50-52`), so it is not a safe substitute. My own captured output shows `Checkpoint committed: .../step_10 (step=10)` and `Training complete: 10 steps, 81,920 tokens`, i.e. the checkpoint is written and the loop completes; only the expected substring is outdated.
2. **All `pp > 1` runs are broken on `main`** (#133, category B — this is that category's first root cause). Running the smoke `test_dense_pp2` command directly (bypassing pytest, which lets torchrun swallow the child traceback) gives the same `AttributeError: 'PipelineStageModule' object has no attribute 'cross_attention_layers'` on both trees, from the same frame in code this PR does not touch: `apply_fsdp2` -> `_fsdp_wrap_transformer_blocks` (`kempnerforge/distributed/parallel.py:239`), which iterates `transformer.cross_attention_layers` — defined on `Transformer` but not on `PipelineStageModule`. `pytest -k test_dense_pp2 --smoke` also fails on the clean `main` baseline. This explains `test_pipeline_parallel`, `test_pp_checkpoint_save_and_resume` and smoke `test_dense_pp2` together.

3. **A `torch.compile` flake in the smoke tier, disclosed rather than smoothed over.** The ID-level comparison flagged `test_dense_compile` failing on clean `main` but **passing** on this branch. An improvement in that direction is not good news — it would imply nondeterminism on that path — so I held the PR and reproduced it: the exact command (`debug.toml --train.compile_model=true`, 4 GPUs) run directly, three times per tree, interleaved in one certified allocation, gave **main 3/3 complete, branch 3/3 complete**. So `main`'s failure was a one-off, not a tree difference, and `tests/smoke/` is not a fully reliable attribution instrument on this path.

   Scope: this cannot reach the parity, resume or DCP evidence, because none of those legs compile. All three parity configs set `compile_model = false`, explicitly overriding the `TrainConfig` default of `True`, and the logs corroborate it — **both** compile messages (`"Compiling model with torch.compile"` from `parallel.py:591` for the dense/MoE path, and `"Compiling VLM wrapper with torch.compile"` from `:487` for the VLM path) appear zero times in all nine parity runs. `--train.compile_model=true` is set only by the smoke test. Determinism of the uncompiled path is separately evidenced by its own three `main`-vs-`main` noise floors coming back bit-exact.

## Testing
- [x] `uv run pytest tests/unit/ -v --timeout=60` passes (1733 tests, 3 skipped, 0 failures)
- [x] `uv run pytest tests/integration/ -v` — 77 passed, 2 skipped (1xH200)
- [x] `uv run torchrun --nproc_per_node=4 -m pytest tests/distributed/ -v` — 99 passed, 2 skipped (all 4 ranks)
- [x] `uv run pytest tests/e2e/ --e2e -v` — 5 failed, 26 passed, 1 skipped. Compared against clean `main` in the same allocation **by test id, both directions**: 32 ids covered on each side, identical status for every one, and the same five failures. No test fails here that passes on `main`, and none passes here that fails on `main`.
- [x] `uv run pytest tests/smoke/ --smoke -v` — 1 failed, 15 passed, 3 skipped (`main` in the same allocation: 2 failed, 14 passed, 3 skipped). The shared failure is `test_dense_pp2`; `main`'s extra failure was a flake, see below.
- [x] `uv run sphinx-build -W --keep-going -b html docs docs/_build/html` — build succeeded
- [x] Tested on 4 GPUs with configs: `debug.toml`, `debug_moe.toml`, `vlm_debug.toml`; and on **2 nodes x 4 GPUs** (`world_size=8`) with `debug.toml` via `scripts/slurm/multinode.sh` — trained 10 steps / 163,840 tokens, committed a checkpoint, then auto-resumed in a second srun job: `Resumed DataLoader: epoch=1, skip_batches=3`, `Starting training: step=10`, `Training complete: 20 steps, 327,680 tokens`

## A note on the certified numbers
Every GPU number above was measured against commit `8aaa72b`, in SLURM allocations that hard-fail unless `git rev-parse HEAD` matches that hash with an empty `git status --porcelain` on both this branch and a clean `origin/main` worktree. Since then this branch has taken test-only commits, a merge of `origin/main`, and the review-requested naming changes. `git diff --stat 8aaa72b..HEAD -- kempnerforge/ scripts/` lists six files: `config/job.py`, `config/loader.py` and `training/optimizer.py` come from the merged #175 and #183 and are not part of this refactor; the remaining three are the naming change. That change does not alter what the loop executes — `training/loop.py`'s entire diff against `8aaa72b` is seven lines, every one of them the `_clip_grads` / `_add_moe_losses` identifier, while `training/__init__.py` and `scripts/train.py` only re-export `run_training` and import it by its package path. The numbers still describe the code they were taken from.

Closes #130
Closes #181


amazloumi:
On the `codecov/patch` failure (58.7%, 298 lines) — this is an artifact of what the refactor moves, not a coverage regression.

`pyproject.toml` sets `[tool.coverage.run] source = ["kempnerforge"]`, on this branch and on `main` alike. `scripts/train.py` has therefore never been measured. This PR moves ~1000 lines out of it (1087 → 42) into `kempnerforge/training/`, so code that previously had **zero** measured coverage becomes visible to the tool for the first time and is counted as newly added. Covered lines strictly increase; the patch percentage looks poor only because the denominator is code that was previously exempt by construction.

Per-module, the uncovered lines are the paths a CPU runner cannot reach:

| module | patch cov | what is uncovered |
|---|---|---|
| `freeze.py` | 100% | — |
| `runtime.py` | 65% | `setup_distributed` (needs `cuda.set_device`) |
| `loop.py` | 63% | `pipeline_step` (PP, 4 GPUs), `vlm_step` (VLM model) |
| `data_pipeline.py` | 57% | VLM/video builders, HF streaming, eval broadcast |
| `entry.py` | 43% | `build_model`, the PP checkpoint group, resume's VLM/MoT branches |

Every one of those is exercised by the GPU tiers reported in the Testing section — integration, distributed on 4 ranks, e2e, smoke, and the parity/resume/2-node legs. `codecov` never observes them because `.github/workflows/ci.yml` gates `gpu-tests` on `workflow_dispatch`, so only `tests/unit/` contributes to the measured run.

The genuinely new logic that *is* CPU-testable carries unit coverage: `tests/unit/test_train_entry.py` adds 51 tests against a fake `CheckpointManager`, taking the suite from 1682 to 1733.

kempnerforge/training/loop.py
```diff
@@ -0,0 +1,673 @@
...
+# Step bodies
+# ---------------------------------------------------------------------------
+
+
+def _clip_grads(model: torch.nn.Module, max_norm: float) -> float:
+    grad_norm = clip_grad_norm_(model, max_norm)
+    return grad_norm.item() if isinstance(grad_norm, torch.Tensor) else grad_norm
+
+
+def _add_moe_losses(loss: torch.Tensor, model: torch.nn.Module, mc: Any) -> torch.Tensor:
```
Naeemkh: Why this function is private? I would suggest dropping the `_` from the function name.

kempnerforge/training/loop.py
```diff
@@ -0,0 +1,673 @@
...
+            self._batch_stream = BatchStream(self.data)
+        return self._batch_stream
+
+
+# ---------------------------------------------------------------------------
+# Step bodies
+# ---------------------------------------------------------------------------
+
+
+def _clip_grads(model: torch.nn.Module, max_norm: float) -> float:
```
Naeemkh: Why this function is private? I would suggest dropping the `_` from the function name.

scripts/train.py
```diff
@@ -22,1065 +25,17 @@
...
-)
-from kempnerforge.training.freeze import (
-    apply_freeze_specs,
-    canonical_freeze_meta,
-    effective_freeze,
-)
-from kempnerforge.training.hooks import HookRunner, StepContext
-
-logger = get_logger(__name__)
+from kempnerforge.training.entry import run_training
```
Naeemkh: Not sure if importing from submodule is a good idea. 

How about using the following:

```
from kempnerforge.training import run_training
```

And adding `run_training` (and any other public API in submodules) into `__all__`.

Naeemkh (dismissed):
Looks good to me. I left a couple of inline comments, both about naming. Nothing that changes behavior.

One thing for later, not for this PR. A researcher should be able to put a training run together from the pieces here without going through `run_training`. Could you open an issue to add an example under `examples/` that does that, and writes down the order the steps have to happen in?

The reason I care about the order being written down is that getting it wrong does not crash. If you build the dataloader before restoring the checkpoint, the run resumes and then trains again on batches it already saw. It logs Resumed from step 6 and exits 0, so it looks fine. The only sign that anything went wrong is a log line that is missing.

amazloumi:
@Naeemkh, Applied the requested changes. The PR needs to be re-approved.

Naeemkh (approved):

PR #183, opened 2026-08-25
Forward zero_grad to Muon's inner AdamW

### LINKED ISSUE #182: Muon never zeroes the gradients it delegates to its inner AdamW
Muon splits its parameters at [`training/optimizer.py:508-511`](https://github.com/KempnerInstitute/KempnerForge/blob/17248f87347b4793727c6ea9aaac4fda103e8d21/kempnerforge/training/optimizer.py#L508), inside _build_muon: anything _is_muon_eligible accepts goes to Muon, the rest to an internal torch.optim.AdamW. Only the Muon half is then registered with the [base class](https://github.com/KempnerInstitute/KempnerForge/blob/17248f87347b4793727c6ea9aaac4fda103e8d21/kempnerforge/training/optimizer.py#L395):

          super().__init__(muon_params, defaults)

`torch.optim.Optimizer.zero_grad` walks `self.param_groups`, which therefore holds only the Muon half. It has no way to know self._adam exists. Muon bridges that gap everywhere the delegation was written out by hand, and `zero_grad` is the one place it was not.

Naeemkh:
## Summary

  `Muon` registers only its own half of the parameters with the base class
  (`optimizer.py:395`), so the inherited `zero_grad` never reached anything
  delegated to the inner AdamW and those gradients accumulated for the entire run,
  measured at 1x, 2x, 3x across three steps with the learning rate held near zero.
  The routing rule sends embeddings and output heads to AdamW on an aspect ratio
  above 10, so at the `configs/train/7b_16gpu_muon.toml` shape 1,050,939,392
  parameters were affected, roughly 13% of the model including its two largest
  tensors, silently and with no warning. `state_dict`, `load_state_dict` and `step`
  already delegate to `self._adam`; this adds the one missing forward plus three
  tests. The routing is shape-dependent, so the tests use a vocab-to-dim ratio
  above 10, because at a lower ratio the embedding stays on Muon's side and they
  pass on the unfixed tree.

## Testing
- [ ] If distributed code changed: `uv run torchrun --nproc_per_node=4 -m pytest tests/distributed/ -v`

Closes #182

### DIFF
--- a/kempnerforge/training/optimizer.py
+++ b/kempnerforge/training/optimizer.py
@@ -412,6 +412,16 @@ class Muon(torch.optim.Optimizer):
         if adam_state is not None and self._adam is not None:
             self._adam.load_state_dict(adam_state)
 
+    def zero_grad(self, set_to_none: bool = True) -> None:
+        """Zero both Muon's own parameters and the inner AdamW's.
+
+        Only the Muon half is registered with the base class, so the inherited
+        ``zero_grad`` would leave every delegated parameter accumulating.
+        """
+        super().zero_grad(set_to_none=set_to_none)
+        if self._adam is not None:
+            self._adam.zero_grad(set_to_none=set_to_none)
+
     @torch.no_grad()
     def step(self, closure=None):
         loss = None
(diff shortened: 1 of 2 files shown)


amazloumi (approved):
LGTM

PR #184, opened 2026-08-25
Move VLM evaluation into examples/vlm/eval

### LINKED ISSUE #143: Migrate vlm evaluation pipeline to examples
## What
Migrate the VLM evaluation harness and script to `examples/vlm-evaluation/` to keep the kempnerforge core focused.

## Scope
This will refactor and migrate `scripts/vlm_eval_harness.py` and `kempnerforge/eval` as well as relevant tests. 

## Backward compatibility
Fully backwards compatible.

amazloumi:
## Summary

- Move the VLM evaluation pipeline to a standalone `examples/vlm/eval/` (`adapter.py`, `vlm_eval_harness.py`, `tests/{unit,integration}/`, and the how-to page as its `README.md`), all via `git mv`. `kempnerforge/eval/` is deleted — no re-export, no shim.
- **BREAKING:** drop `[project.entry-points."lmms_eval.models"] kempnerforge_vlm` from `pyproject.toml` and its `manifest.py` (`examples/` is not part of the installed distribution, so a manifest there can never be discovered). lmms-eval's own CLI no longer resolves `kempnerforge_vlm`; run the harness by path, flags unchanged:
  ```bash
  uv run python examples/vlm/eval/vlm_eval_harness.py \
      --config configs/train/vlm_jd.toml --checkpoint checkpoints/vlm/step_10000 --tasks mmmu_val
  ```
- Replacing that discovery: the harness builds `KempnerForgeVLM` itself and passes the instance to `simple_evaluate`, importing `adapter` as a `sys.path[0]` sibling module; the tests reach it the same way via the example's `tests/conftest.py`.
- Contract tests swap targets: the entry-point and manifest pins are gone, replaced by `TestSimpleEvaluateContract`, which pins the exact `simple_evaluate` call the harness now depends on.
- The example's tests are standalone and must run as **two separate pytest sessions** — the unit conftest's hermetic fake `lmms_eval` would otherwise replace `adapter` under the integration tests. The README documents this.
- Also updated: `CHANGELOG.md` (BREAKING entry), `docs/how-to/index.md` (links to the example), `kempnerforge/config/vlm.py` docstring, and the harness docstring's stale `v1 is single-GPU and image-only` → `single-GPU` (video has been supported since the video PRs).
- The example's tests now live outside `tests/`, so PR CI no longer runs them or counts them in codecov — a repo-wide call for all examples rather than something to settle here.

## Testing

- [x] `uv run ruff format --check kempnerforge/ tests/ scripts/` passes (162 files)
- [x] `uv run pytest tests/unit/ -q --timeout=180` passes (1592 passed, 3 skipped)
- [x] `uv run pytest examples/vlm/eval/tests/unit -q` passes (104 passed)
- [x] `pytest examples/vlm/eval/tests/integration -v` **with real lmms-eval 0.7.2 installed**: 7 passed, 1 skipped (the env-gated real-checkpoint test). All five contract classes executed — `ChatMessages`, `Collator`, `simple_evaluate`, the `lmms` base, and `Instance` — so the unit tests' fake is confirmed against the real API, and the new instance-path call shape holds.
- [x] **Real harness run on an H200** (`kempner_eng`), launched from `/tmp` where `adapter` is not importable from cwd or `PYTHONPATH` (asserted in-job): `vlm_eval_harness.py --tasks mlvu_dev --limit 2` against real lmms-eval 0.7.2 and real MLVU-dev videos exits 0 — the `sys.path[0]` sibling import resolves, the DCP checkpoint loads (888 tensors), `generate_until` decodes both requests on GPU, and `simple_evaluate` returns results JSON with `config.model_args` recorded. The score is 0.0 because that checkpoint carries untrained weights: every trained VLM checkpoint on the cluster is Qwen3-0.6B geometry (`q_proj [2048, 1024]`, i.e. head_dim 128), which `main` cannot express — `ModelConfig.head_dim` is a read-only `dim // n_heads` — so it fails at `dcp.load` with a size mismatch on `main` with or without this PR.
- [x] `uv run pytest tests/integration/ --collect-only -q` → 77 collected; `tests/conftest.py` is unmodified, so the fixtures the core suite shares stay where they were
- [x] `grep -rnE '^\s*(from|import)\s+(examples|adapter)\b' kempnerforge/ scripts/ configs/` → no output: no core module imports the example
- [x] `grep -rIn 'kempnerforge\.eval|kempnerforge/eval|scripts/vlm_eval_harness|kempnerforge_vlm|run-vlm-evaluation|tests/unit/eval' --exclude=CHANGELOG.md .` → no output; `CHANGELOG.md` keeps the historical `### Added` entry plus the new BREAKING note that supersedes it
- [x] `uv run sphinx-build -W --keep-going -b html docs docs/_build/html` succeeds (the docs job triggers on this diff)
- [ ] Distributed / e2e: not applicable — file move, no training-loop, parallelism, or numerics change

Closes #143


Naeemkh (approved):
Looks good to me. 
One ask before this lands: add `uv run pytest examples/vlm/eval/tests/unit -q to ci.yml`. `tests/unit/eval/` collects 105 tests on main, and after this nothing runs them automatically. Your README says the unit tier is hermetic, so it needs no new dependency, and putting the invocation in the workflow also pins the two-separate-sessions requirement somewhere that executes rather than in prose. Integration can stay manual since it needs real lmms-eval.

amazloumi:
Thanks. will have a followup PR for the requested CI change.

PR #187, opened 2026-08-26
Guard the cross-attention FSDP2 wrap for pipeline stage modules

### LINKED ISSUE #185: Pipeline parallelism crashes on startup
## What happened

  Any run with `pp > 1` dies before the first training step.

## Reproduction

  Reproduced on 4x H200:

```
uv run torchrun --nproc_per_node=4 scripts/train.py configs/train/debug.toml
      --distributed.pp=2 --distributed.dp_shard=2 --train.grad_accum_steps=2 --train.max_steps=10
```

All four ranks fail with the same error:

```
AttributeError: 'PipelineStageModule' object has no attribute 'cross_attention_layers'
    scripts/train.py:144         in main
    distributed/parallel.py:288  in apply_fsdp2
    distributed/parallel.py:239  in _fsdp_wrap_transformer_blocks
```

## Error / traceback

  `parallel.py:239` reads an attribute without checking that it exists:

```python
  for ca_block in transformer.cross_attention_layers.values():
```
  Every Transformer has cross_attention_layers (empty for non cross-attention
  configs), so this line is correct for normal runs. Under pipeline parallelism the
  object passed in is a PipelineStageModule, which does not define it, so the
  lookup raises.

  The parameter is annotated transformer: Transformer, but scripts/train.py
  passes a PipelineStageModule at both :133 and :144. That mismatch predates
  the crash. It was harmless while the function only touched .layers, which both
  classes happen to define, and became fatal when the line above was added.

Naeemkh:
## Summary

`apply_fsdp2` receives a `PipelineStageModule` under pipeline parallelism, but _fsdp_wrap_transformer_blocks read transformer.cross_attention_layers unconditionally. Only Transformer defines that attribute, so every pp > 1 run died with AttributeError before the first step.

- Guard the cross-attention FSDP2 wrap with `getattr`. `apply_fsdp2` receives a `PipelineStageModule` under PP, which has no `cross_attention_layers`, so every `pp > 1` run died with `AttributeError` before the first step.
- Widen both signatures to `Transformer | PipelineStageModule`. With the annotation widened and the guard removed, `pyright` flags the offending line, so CI catches the next one.
- Add the first test that calls `apply_fsdp2` on a stage module. CPU only, runs in CI today, fails on the unpatched tree.

## Testing
- [ ] If training loop / parallelism / optimizers changed: `uv run pytest tests/e2e/ --e2e -v`

Closes #185


kempnerforge/distributed/parallel.py
```diff
@@ -234,9 +239,10 @@ def _fsdp_wrap_transformer_blocks(
             reshard_after_forward=reshard_after_forward,
         )
 
-    # Cross-Attention blocks (Cross-Attention arch only). Empty dict
-    # for non-CA configs.
-    for ca_block in transformer.cross_attention_layers.values():
+    # Cross-Attention blocks (Cross-Attention arch only). Empty dict for
+    # non-CA configs, and absent entirely on a PipelineStageModule, which
+    # defines no cross_attention_layers.
+    for ca_block in getattr(transformer, "cross_attention_layers", {}).values():
```
amazloumi, line 247: `getattr` with a string literal is invisible to pyright, so widening the annotation buys no checking on this line.

tests/unit/test_distributed.py
```diff
@@ -508,3 +508,77 @@ def fake_fully_shard(mod, **kwargs):  # noqa: ARG001
         # Each captured object should be a TransformerBlock, not a sub-module.
         for layer in captured:
             assert isinstance(layer, TransformerBlock)
+
+    def test_pipeline_stage_module_gets_block_level_wrap(self, monkeypatch):
```
amazloumi: is this test covering `_fsdp_wrap_transformer_blocks` or `apply_fsdp2` on a stage module. it seems the `fully_shard(model)` stay uncovered. better to double check if this test works the way it is intended to.

kempnerforge/distributed/parallel.py
```diff
@@ -182,7 +187,7 @@ def _has_ep_moe(module: torch.nn.Module) -> bool:
 
 
 def _fsdp_wrap_transformer_blocks(
-    transformer: Transformer,
+    transformer: Transformer | PipelineStageModule,
```
amazloumi, line 190: the docstring below needs to be updated similar to updated inline comment at lines 242-245.

amazloumi (dismissed):
LGTM. none of comments are blockers.

kempnerforge/distributed/parallel.py
Naeemkh: You're right that `pyright` can't see through `getattr` with a string literal. However, the widening isn't aimed at that line. That line is guarded, so it's safe by construction. It's there so the next bare attribute access on the parameter gets flagged, and so the call sites stop being type violations. With the annotation widened and the guard removed, pyright reports 3 errors on this line; under the old Transformer-only annotation it reported zero, because the declaration was a lie that happened to typecheck.

tests/unit/test_distributed.py
Naeemkh: Good catch, it was hitting the helper not `apply_fsdp2`, so the top-level wrap was uncovered. I fixed it.

kempnerforge/distributed/parallel.py
Naeemkh: Done, the docstring now names all three states of cross_attention_layers instead of only the empty one.

amazloumi (approved):
Thanks for the changes.

PR #190, opened 2026-08-26
Move the VLM COCO prep helper into examples/vlm/

### LINKED ISSUE #189: Move the VLM COCO prep helper into examples/vlm/
## What
`scripts/prep_vlm_coco.py` builds a COCO-Karpathy HuggingFace dataset on disk for the VLM experiment. That is experiment-specific data prep, so it belongs with the example — core `scripts/` should hold only general tooling.

## Scope
- `git mv scripts/prep_vlm_coco.py examples/vlm/data/prep_vlm_coco.py`.
- Update the script's own usage strings (module docstring and argparse epilog) so they print the new invocation path.
- Drop its `codebase-map.json` entry; the map does not track `examples/`.
- Drop the two cluster-specific absolute paths it carries as `--caption-json` / `--image-root` defaults, so the shipped example is general.

## Backward compatibility
**BREAKING** for anyone invoking `scripts/prep_vlm_coco.py`. No compat shim or symlink is added; the new path is `examples/vlm/data/prep_vlm_coco.py`. `--caption-json` and `--image-root` become required, because their previous defaults were cluster-specific absolute paths that a shipped example must not carry. `--num-samples` still defaults to `500`, so a smoke prep stays a short command.

amazloumi:
## Summary
- Move `scripts/prep_vlm_coco.py` to `examples/vlm/data/prep_vlm_coco.py`: it is VLM-experiment data prep, not general core tooling.
- Update the script's own usage strings and drop its `codebase-map.json` entry (the map tracks core only).
- Drop the two cluster-specific absolute paths it carried as `--caption-json` / `--image-root` defaults; both flags are now required, so the shipped example is machine-independent.
- Point the `vlm_7b_siglip2.toml` comment at the new path.
- **BREAKING** on both the script path and the now-required flags; no compat shim.

## Testing
- [x] `uv run ruff format --check kempnerforge/ tests/ scripts/` passes — 180 files
- [x] `uv run pytest tests/unit/ -v --timeout=120` — 1828 passed, 3 skipped
- [x] `git grep -n prep_vlm_coco` returns only the new path and CHANGELOG history
- [x] Ran the moved script from its new path: 500-sample single split and `--all-splits`; reloaded both with `load_from_disk` and checked columns, image decode, and that val/test rows keep 5 `references` with `caption` still a single `str`; plus the `--help`, missing-required-arg, existing-`--out`, and missing-sibling-JSON branches
- [ ] If distributed code changed: n/a — single-process data-prep script, no collectives, checkpointing, or parallelism
- [ ] If training loop / parallelism / optimizers changed: n/a

Closes #189


Naeemkh:
@amazloumi, the branch is out of date, please update.

Naeemkh (approved):
Looks good to me.

PR #192, opened 2026-08-26
Run the examples/ unit test tier in CI

### LINKED ISSUE #191: Run the examples/ unit test tiers in CI
## What
Tests that move out of `tests/` into `examples/*/tests/` stop running in CI. The `unit-tests` job invokes only `pytest tests/unit/`, and `[tool.pytest.ini_options] testpaths = ["tests"]` means no default collection ever reaches `examples/`. The relocated VLM-eval unit tier (104 tests) is currently unexercised on every push and PR.

## Scope
- Add an explicit `pytest examples/vlm/eval/tests/unit` step to the `unit-tests` job in `.github/workflows/ci.yml`.
- One pytest session per tier: the VLM-eval unit conftest injects a hermetic fake `lmms_eval` into `sys.modules`, which would leak into any tier sharing the session.
- The matching integration tier needs the real `lmms-eval` — an intentionally undeclared optional dependency — so it stays a manual run.
- Update the CI job table in `CONTRIBUTING.md`.
- Lint scope stays `kempnerforge/ tests/ scripts/`; whether ruff/pyright should cover `examples/` is a separate decision.

## Backward compatibility
CI-only, no runtime change. The added step is a new gate over code that was already expected to pass.

amazloumi:
## Summary
- New `ci-example.yml` runs the hermetic example test tiers, which sit outside the root `testpaths` and so were collected by no CI step after #184. One job per example behind a `dorny/paths-filter` gate, so touching one example does not run another's tests. Today that is `vlm-tests` running `pytest examples/vlm/eval/tests/unit`; a second tier in the same example is one more `- run:` line.
- Both triggers are path-filtered to `examples/vlm/**`. The workflow's own path is deliberately not listed: it matched on every edit to the file, so adding an example would have run every example's tests.
- One pytest session per tier — the tier's conftest injects a hermetic fake `lmms_eval` that would leak into any tier sharing the session. The VLM-eval integration tier needs the real, undeclared `lmms-eval` and stays manual.
- `ci.yml` is byte-identical to `main`; ruff and pyright scope unchanged.
- Adds `dorny/paths-filter@v3` — the first action used here outside `actions/`, `astral-sh/` and `codecov/`.
- Consequence: with both triggers filtered, a core change that breaks an example tier is not caught by this workflow; it surfaces when the tier next runs. The VLM-eval adapter has 10 imports from `kempnerforge`.
- Consequence: the ruleset's required contexts are `lint` and `unit-tests`. A path-filtered check cannot be made required — GitHub blocks every PR where it never reports — so these jobs are advisory unless someone adds them to the ruleset and accepts that.
- Open question, not settled here: `pyproject.toml` and `uv.lock` are not in the filter, so a dependency bump can break an example tier without touching any example directory. "Run every tier" is a different rule from per-example filtering, so it seems worth deciding separately.

## Testing
- [x] `uv run pytest tests/unit/ -v --timeout=120` on the final tree (`2438705`) — 1726 passed, 3 skipped in 237.76s
- [x] The tier in a venv proven free of lmms-eval — `find_spec("lmms_eval")` is `None`, `import lmms_eval` raises `ModuleNotFoundError`, and no orphaned `lmms_eval/` directory is on `sys.path`. `find_spec` alone is not sufficient: `uv sync` prunes the package's modules but leaves its task-data files, and that `__init__.py`-less directory reads as a namespace package. Result: 104 passed, exit 0.
- [x] On the runner, `changes` resolved `dorny/paths-filter@v3` and logged `Filter vlm = true` / `Changes output set to ["vlm"]`; the gated `vlm-tests` job then reported `collected 104 items` / `104 passed in 6.78s`.
- [x] Skip case observed, which the earlier filter could not show: `b25ce48` touched only `.github/workflows/ci-example.yml`, `CHANGELOG.md` and `CONTRIBUTING.md` — `CI` ran and `CI (examples)` produced no run at all. `2438705` then added `examples/vlm/eval/README.md` and `CI (examples)` ran and passed.
- [x] The job can fail: reversing the comparison in `adapter.py::_first_stop`, so it returns the last stop match instead of the first, gives `1 failed, 103 passed` and exit 1. Defect reverted.
- [x] Test path is `examples/vlm/eval/tests/unit`, which exists on this base. `examples/vlm/tests` (added by #193) does not, and naming it would exit 4 — confirmed by running it.
- [x] `uv run ruff check kempnerforge/ tests/ scripts/` — All checks passed. `uv run ruff format --check kempnerforge/ tests/ scripts/` — 166 files already formatted.
- [x] `uv run pyright kempnerforge/` — 0 errors, 0 warnings, 0 informations.
- [x] `ci-example.yml` parses to jobs `changes` + `vlm-tests`, with `permissions: contents/pull-requests: read`; `ci.yml` is byte-identical to `main`.
- [ ] Distributed / fault-tolerance / resumability battery deliberately not run: the diff is workflow YAML plus docs and touches no runtime code, so there is no path by which it can affect those guarantees.

Closes #191


.github/workflows/ci.yml
```diff
@@ -26,6 +26,11 @@ jobs:
       - uses: astral-sh/setup-uv@v6
```
Naeemkh, line 26: Please create ci-example.yml file, and trigger the tests only if the corresponding examle folder files are modified.

Naeemkh (changes requested):

.github/workflows/ci.yml
amazloumi: done

.github/workflows/ci-example.yml
```diff
@@ -0,0 +1,29 @@
...
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+    # Scoped to `examples/` and its deps so the filter actually filters. Core is
+    # deliberately not listed: the unfiltered `push` above is what catches a core
+    # change that breaks an example tier without touching `examples/`.
+    paths:
+      - "examples/**"
```
Naeemkh: This includes all examples, we have to have one workflow dispatch per example. So, it should be e.g., "examples/vlm". 

And later on you need to be able to filter the workflow based on those.  I came up with the following not sure, if it is the best option. If it takes too much to figure this out, just add vlm to your pattern, we can circle back to this later. 

```
name: CI (examples)

on:
  push:
    branches: [main]
    paths:
      - "examples/example-one/**"
      - "examples/example-two/**"

  pull_request:
    branches: [main]
    paths:
      - "examples/example-one/**"
      - "examples/example-two/**"

  workflow_dispatch:

permissions:
  contents: read
  pull-requests: read

jobs:
  changes:
    runs-on: ubuntu-latest
    outputs:
      example_one: ${{ steps.filter.outputs.example_one }}
      example_two: ${{ steps.filter.outputs.example_two }}

    steps:
      - uses: actions/checkout@v5

      - uses: dorny/paths-filter@v3
        id: filter
        with:
          filters: |
            example_one:
              - "examples/example-one/**"

            example_two:
              - "examples/example-two/**"

  example-one-tests:
    needs: changes
    if: needs.changes.outputs.example_one == 'true' || github.event_name == 'workflow_dispatch'
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v5
      - uses: astral-sh/setup-uv@v6
      - run: uv sync --group video
      - run: uv run pytest examples/example-one/tests -v --timeout=60

  example-two-tests:
    needs: changes
    if: needs.changes.outputs.example_two == 'true' || github.event_name == 'workflow_dispatch'
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v5
      - uses: astral-sh/setup-uv@v6
      - run: uv sync --group video
      - run: uv run pytest examples/example-two/tests -v --timeout=60
```

.github/workflows/ci-example.yml
```diff
@@ -0,0 +1,29 @@
...
+      - "pyproject.toml"
+      - "uv.lock"
+      - ".github/workflows/ci-example.yml"
```
Naeemkh: This file will be changed as we  add new example, so we don't want to trigger other examples tests. So, let's drop it.

Naeemkh (approved):
Thank you.

PR #198, opened 2026-09-02
Config-driven model knobs: decoupled attention head dim, adapter pre-norm, weights-only warm-start

### LINKED ISSUE #197: Config-driven model knobs: decoupled attention head dim, adapter pre-norm, weights-only warm-start
## What

Three general knobs are missing from core.

**1. Attention head width is hard-coupled to `dim // n_heads`.** `ModelConfig.head_dim` is a computed property, so an architecture whose per-head width is independent of the residual width (`n_heads * head_dim != dim`) cannot be expressed at all. Everything downstream already reads `config.head_dim` — attention, RoPE precompute, KV-cache sizing, the FLOP/MFU model, `num_params_estimate` — and `Attention`, `MoTAttention` and `CrossAttention` already accept a `head_dim` argument. Only the config-level override and the `CrossAttentionBlock` pass-through are absent.

**2. `mlp_2layer` has no pre-projection norm.** A connector that normalizes vision features before the first projection currently requires a whole new adapter type, even though the difference is one norm chosen from the existing norm registry.

**3. `checkpoint.exclude_from_loading` is declared but never read.** Nothing passes it to `CheckpointManager.load(exclude_keys=...)`, so a weights-only checkpoint — a converted external model with no optimizer or train state — cannot be warm-started cleanly. The field's documented meaning ("FQN prefixes to skip on load") also does not match what `load` accepts, which is DCP state keys (`"model"` / `"optimizer"`).

## Scope

- `ModelConfig`: an override for the per-head width, `0` meaning "infer `dim // n_heads`". `dim % n_heads == 0` is then required only when inferred; a negative override is rejected at config time.
- Keep the resolved head width **derived, not stored**. `load_config` overlays TOML and CLI values onto an already-constructed default config and carries every unspecified field forward, so a value resolved into a field in `__post_init__` goes stale as soon as `dim` or `n_heads` is set from TOML — silently changing the architecture rather than failing.
- `CrossAttentionBlock`: forward `head_dim` to `CrossAttention`, but only when the override is set. Cross-attention has its own head count (`vlm.cross_attention_n_heads`); forwarding a coupled `dim // n_heads` unconditionally would resize every cross-attention block whenever the two counts differ.
- `AdapterConfig.pre_norm`: a norm-registry key for a norm over the vision features before `mlp_2layer`'s first projection, validated against the registry; `""` disables it. This keeps a model-specific adapter type out of core.
- Honor `checkpoint.exclude_from_loading` on a `load_path` warm start only; a resume must ignore it and always restore full state. Correct the field's documentation.

## Backward compatibility

All three default to today's behavior: the override at `0` reproduces `dim // n_heads`, `pre_norm=""` builds no module and leaves the state dict and RNG stream untouched, and an empty `exclude_from_loading` passes `None`.

amazloumi:
## Summary

- `model.head_dim_override` (default `0` → `dim // n_heads`) decouples the attention head width from the residual width, so `n_heads * head_dim` need not equal `dim`. `dim % n_heads == 0` is now required only when the head width is inferred; a negative override is rejected at config time.
- `head_dim` stays a derived property rather than a resolved field. `load_config` overlays TOML and CLI values onto an already-constructed default config and carries unspecified fields forward, so a stored value would go stale as soon as `dim` or `n_heads` came from TOML.
- `CrossAttentionBlock` forwards `head_dim` to `CrossAttention`, but only when the override is set. Cross-attention has its own head count (`vlm.cross_attention_n_heads`), so under the default coupling it keeps its existing `dim // n_heads` width.
- `adapter.pre_norm` applies a norm-registry-selected norm to the vision features before `mlp_2layer`'s first projection, exposed as `ln_q`; `""` (default) builds no module. This is what keeps a model-specific adapter type out of core.
- `checkpoint.exclude_from_loading` is now honored when warm-starting from `checkpoint.load_path` and ignored on a resume, which always restores full state. It was previously declared but never read, and its docs described FQN prefixes rather than the DCP state keys `CheckpointManager.load` accepts.
- Overlaps other in-flight `examples/`-realignment work in `kempnerforge/model/adapter.py`, `kempnerforge/config/adapter.py` and `tests/unit/test_adapter.py` (ragged `attentional_pool`), plus `CHANGELOG.md`; whichever lands second will conflict in those files.

## Testing

- [x] `uv run ruff format --check kempnerforge/ tests/ scripts/` passes — 166 files
- [x] `uv run pytest tests/unit/ -v --timeout=120` — 1785 passed, 3 skipped (+58 new)
- [x] `uv run pytest tests/integration -q` (4 GPUs) — 77 passed
- [x] 4-GPU `tests/distributed/` in full — 99 passed, 2 skipped, on each of 4 ranks (`test_checkpoint.py` + `test_resilience.py` alone: 13 passed per rank)
- [x] **Defaults unchanged**: `dim=256, n_heads=4, n_kv_heads=2`, 4 GPUs, seed 42, pre-tokenized data, 40 steps. Two runs of clean `origin/main` (separate checkout, own venv) are bit-identical to each other, and this branch is bit-identical to both at all 40 logged steps — step 1 `10.4375`, step 40 `9.375`. Losses read at full precision from TensorBoard event files, not the 4-dp console line. The comparator refuses instead of reporting equality when a run is missing, empty, truncated, step-shifted or non-finite, or when the control pair itself disagrees; all 9 refusal/power cases self-tested in the same job.
- [x] **Decoupled `head_dim`, 4-GPU FSDP2**: `head_dim_override=96` with `n_heads=4, dim=256` (`4 * 96 = 384 != 256`), 19,925,248 params — loss `10.4375 → 10.0625` over 25 steps, `grad_norm 0.961 → 1.938`.
- [x] **Decoupled `head_dim`, 2 nodes × 4 GPUs** via `scripts/slurm/multinode.sh`, FSDP2 over `dp_shard=8` — loss `10.4375 → 9.5625` over 30 steps, `grad_norm 0.664 → 1.617`, 1.03M tok/s, no wedge or rank divergence. A second 2-node run under `tp=2 × dp_shard=4` also completes 30 steps cleanly.
- [x] **TP is unaffected by this change.** Clean `origin/main` with a *coupled* `head_dim` is already flat under `tp=2`, at both model sizes, so the flat loss is pre-existing and not this PR's. 4 GPUs, 25 steps, seed 42:

  | tree | head_dim | mesh | step 1 | step 25 | grad_norm 1 → 25 |
  |---|---|---|---|---|---|
  | *dim=256, n_heads=4 (~20M)* | | | | | |
  | `origin/main` | 64 coupled | `dp_shard=4` | 10.4375 | 9.8125 | 0.859 → 1.844 |
  | this branch | 64 coupled | `dp_shard=4` | 10.4375 | 9.8125 | 0.859 → 1.844 |
  | `origin/main` (×2 runs) | 64 coupled | `tp=2 × dp_shard=2` | 10.3750 | 10.3750 | 0.006 → 0.015 |
  | this branch | 64 coupled | `tp=2 × dp_shard=2` | 10.3750 | 10.3750 | 0.006 → 0.015 |
  | this branch | 96 **decoupled** | `tp=2 × dp_shard=2` | 10.3750 | 10.3750 | 0.005 → 0.015 |
  | *dim=1024, n_heads=8 (~150M)* | | | | | |
  | `origin/main` | 128 coupled | `dp_shard=4` | 10.5625 | 10.1250 | 5.625 → 6.375 |
  | this branch | 192 **decoupled** | `dp_shard=4` | 10.5625 | 10.4375 | 5.312 → 4.062 |
  | `origin/main` | 128 coupled | `tp=2 × dp_shard=2` | 10.3750 | 10.3750 | 0.011 → 0.013 |
  | this branch | 192 **decoupled** | `tp=2 × dp_shard=2` | 10.3750 | 10.3750 | 0.011 → 0.013 |

  Under `tp=2` with the coupled config, branch and `origin/main` are bit-identical across all 25 steps (comparator PASS, same-code control pair bit-identical). The decoupled head dim trains normally under FSDP2 at both sizes.
- [x] **TP shards from `n_heads * head_dim`, not `dim`**: `apply_tensor_parallel` builds a plan of bare `ColwiseParallel()` / `RowwiseParallel()` and contains no head-width arithmetic, so widths come from each `nn.Linear`'s own `out_features`. Verified under gloo with `dim=256, n_heads=4, n_kv_heads=2, head_dim=96, tp=2`: global `q_proj (384, 256)` → per-rank `(192, 256)`, `k_proj (192, 256)` → `(96, 256)`, `o_proj (256, 384)` → `(256, 192)`. A `dim // tp` shard would be `(128, 256)`. Forward+backward gives finite non-zero grads.
- [x] **SIGTERM preemption drill, twice.** (a) Cooperative path, `dim=1024`/12 layers, 4 GPUs, SIGTERM to the worker ranks at step 300 of 900: `Shutdown requested at step 300 — saving emergency checkpoint`, agent exits 0, `latest -> step_300`, `step=300`, `batches_yielded=300`; resume logs `Resumed from step 300, 2,457,600 tokens seen` (= 300 × 4 × 4 × 512, exact) and `Resumed DataLoader: epoch=0, skip_batches=300`, then runs to 900. Two identical runs of this config are *not* bit-reproducible (same-code control pair max|d| `3.125e-02`), so the comparator refused an exact verdict; against that noise floor the resumed 900-step trajectory differs by max|d| `3.125e-02` — ratio 1.000, i.e. no more than two identical runs differ from each other. (b) Hard-kill path, `dim=256`, SIGTERM to the agent *and* all 4 workers at step 31 of 40: `step_31` was left with zero shards and auto-resume correctly fell back to the newest complete checkpoint `step_30` (`skip_batches=30`, `245,760` tokens, exact); that config *is* bit-reproducible, and the resumed 40-step trajectory is bit-identical to an uninterrupted run.
- [x] **Head-geometry consumer sweep**: every consumer already reads `config.head_dim`, so all follow the override — `Attention` (q/k/v/o widths, `q_norm`/`k_norm`, SDPA scale), `KVCache`, `MoTAttention`, `CrossAttention`, MoMa's `Attention`, RoPE precompute in `Transformer` and `pipeline_parallel`, `generate`'s cache sizing, the dense and MoE FLOP models, `num_params_estimate`. No site assumes `n_heads * head_dim == dim`: every output reshape uses `-1` and `o_proj` maps back to `dim`. Tested per-arch — with the override set, every `Attention`/`MoTAttention`/`CrossAttention` module in a `joint_decoder`, `cross_attention`, `mot` and `moma` build reports the overridden width.
- [x] **Cross-attention regression guard**: a coupled model (`dim=64, n_heads=8`) with `cross_attention_n_heads=2` keeps `head_dim=32` and `q_proj (64, 64)`, verified equal to clean `origin/main`. Forwarding the model's `dim // n_heads` unconditionally would make it `head_dim=8`, `q_proj (16, 64)`.
- [x] **TOML overlay guard**: a TOML setting `dim=256, n_heads=4` with no override resolves `head_dim=64` on both this branch and clean `origin/main`. A `head_dim` resolved into a field in `__post_init__` reads `128` here, carried from the `4096/32` default — silently wrong for `configs/train/debug.toml`, `debug_moe.toml` and `hf_wikitext.toml`.
- [x] **Validation cases**: negative override, indivisible `dim` with no override (and accepted *with* one), unknown `adapter.pre_norm` — all rejected at config time, before distributed init, with the options named.
- [x] **Warm-start/resume asymmetry, both directions**: `exclude_from_loading=["optimizer"]` reaches `load(exclude_keys=...)` on a `load_path` warm start, and is passed as `None` on a resume — including when a stale `load_path` is also set. An empty list passes `None`.
- [x] **`pre_norm` default path**: no `ln_q` module, no `ln_q` state-dict key, and the projections' RNG draw identical to a build without the argument. With it set, `forward` equals `proj2(act(proj1(ln_q(x))))` exactly, and `reset_parameters` restores `weight=1`/`bias=0` after a meta-device `to_empty`.
- [x] `uv run pytest tests/smoke/ --smoke` (4 GPUs) — 16 passed, 3 skipped, no failures; identical counts on clean `origin/main`. Covers `dense`/`moe` × `fsdp`/`tp2_fsdp`, plus compile, FP8, grad-accum, full AC, sigmoid router, `pp2` and inline eval.
- [x] Re-run with a pre-tokenized `--data-path` to unskip the last three: **both `TestAutoResume` tests pass** on this branch — RNG restore, `StatefulDataLoader` replay with monotonic `batches_yielded`, `train_state.pt` ownership gate, scheduler continuity, i.e. the path `restore_checkpoint` sits on.
- [ ] **Open, under investigation:** in that same `--data-path` run, `TestRealConfigs::test_fp8_7b_config` failed on this branch (18 passed, 1 failed) and passed on clean `origin/main` (19 passed), same 4 GPUs, same job. The config resolves byte-identically on both trees (`head_dim=128`, `ffn=14336`, 8,030,261,248 params) and the run is text-only with no resume, so none of the four knobs is on its path — but that is an argument, not evidence. Re-running the test alone and interleaved across both trees to separate a real regression from suite-position / GPU-state effects. Not claiming this one is clean until that says so.
- [x] `uv run pytest tests/e2e --e2e` (4 GPUs) — 5 failed / 26 passed on this branch vs 4 failed / 27 passed on clean `origin/main` in the same session. The 4 shared failures (`test_checkpoint_save_and_resume`, `test_moe_checkpoint_resume`, `test_pp_checkpoint_save_and_resume`, `test_sigterm_triggers_emergency_checkpoint`) are the ones #133 already records as pre-existing on `main`. The one-test delta is `test_hf_dataset_single_gpu`, which #133 records as network-flaky; re-run alone on the branch it passes 3/3, so the branch and `main` fail the same 4 pre-existing tests.
- [x] Re-checked against the design plan's core-admission test at completion: three general model knobs entering by general config fields (`pre_norm` also through the norm registry) plus one correctness fix to an inert config field, defaults unchanged, useful to any decoupled-head-geometry model rather than one experiment.
- [x] If training loop / parallelism / optimizers changed: parallelism application order (TP → EP → FP8 → AC → FSDP2) is untouched — `kempnerforge/distributed/` has no diff.

Closes #197


Naeemkh (changes requested):
Nice work. The head-dim decoupling is the piece people are waiting on, and keeping it a derived property rather than a resolved field is the right call. Building the alternative and measuring it freeze at 128 for the debug configs is much stronger than arguing for it, and the same goes for the cross-attention regression guard.
I would suggest to split this, and let the split follow your own reasoning about the open item. The `fp8` failure is on a text-only config with no resume. That means it cannot implicate the adapter pre-norm (VLM-only) or the warm-start exclusion (neither path runs in that test). Those two are provably outside the question, so they could land now. The head-dim change is the only one even adjacent to it, and it is the one worth holding until the interleaved re-run reports. Right now one unresolved question is gating all four. 
One note on the warm-start change: when `load_path` is set and the checkpoint directory also has content, auto-resume already wins and `load_path` is silently ignored. With this change the exclusion list is silently dropped along with it, so someone asking for a weights-only warm start can get a full resume from a different checkpoint with no warning. Pre-existing, but worth a line in the docs here. Separately, and more important than this PR: your TP row measures clean main sitting at chance under tp=2 with grad_norm around 0.01, at two model sizes, while the data-parallel control learns normally.

PR #203, opened 2026-09-14
Write the emergency checkpoint when dataloader workers are in use

Naeemkh:
## Summary

When SLURM preempts a job it signals every process in the group, not just the main one. That kills the dataloader's worker processes first, so the main process fails trying to fetch its next batch before the shutdown handler ever ran. No emergency checkpoint was written and nothing appeared in the log. With `data.num_workers = 0` the same drill worked, which is what pinned the cause. Workers now ignore the shutdown signals, so the step in progress can finish and the checkpoint is written on the normal path. That has a side effect worth knowing: the usual way to stop a worker is also a signal, so a shielded worker can't be stopped that way and the process would hang on exit waiting for them. They are now shut down explicitly when the training loop ends.

Also, on restart, auto-resume picked the newest checkpoint folder without checking whether it had finished being written. A run whose only checkpoint was a failed emergency save resumed into it and died. It now skips unfinished folders and takes the newest complete one. #194 already made the loader recover from this, but the resume path is also used to read the saved step before loading, and that read doesn't notice an unfinished folder, so resolving to a complete one closes the remaining gap.

## Testing
- [ ] `uv run pytest tests/unit/ -v --timeout=60` passes
- [ ] If distributed code changed: `uv run torchrun --nproc_per_node=4 -m pytest tests/distributed/ -v`
- [ ] If training loop / parallelism / optimizers changed: `uv run pytest tests/e2e/ --e2e -v`

Closes #


Naeemkh:
Closes #177

kempnerforge/training/data_pipeline.py
```diff
@@ -286,6 +287,8 @@ def _build_hf_pipeline(config: JobConfig, dp_rank: int, dp_size: int) -> DataPip
             num_workers=config.data.num_workers,
             pin_memory=config.data.pin_memory,
             prefetch_factor=(config.data.prefetch_factor if config.data.num_workers > 0 else None),
+            # See StatefulDataLoader: workers survive a group-delivered SIGTERM.
+            worker_init_fn=ignore_shutdown_signals_in_worker,
```
amazloumi, line 291: Better to pass the `persistent_workers` true when the number of workers are more than 0. `persistent_workers=config.data.num_workers > 0`

> persistent_workers ([bool](https://docs.python.org/3/library/functions.html#bool), optional) – If True, the data loader will not shut down the worker processes after a dataset has been consumed once. This allows to maintain the workers Dataset instances alive. (default: False)

kempnerforge/resilience/elastic.py
```diff
@@ -17,6 +17,27 @@
...
+    """Whether ``ckpt_dir`` holds a complete set of DCP shards.
+
+    Accepts both layouts: a flat directory, and the per-stage ``pp{k}/``
+    subdirectories written under pipeline parallelism.
+    """
+    if (ckpt_dir / _DCP_METADATA_FILE).exists():
+        return True
+    if not ckpt_dir.is_dir():
+        return False
+    return any(
```
amazloumi, line 35: `any` calls a PP checkpoint durable when only some stages flushed, which is what
  an interrupted PP save looks like (all ranks mkdir their own pp{k}/ up front).
  _resolve_dcp_load_dir then re-tests per rank, so stage 0 resumes at step_20 and
  stage 1 at step_10, silently.

kempnerforge/resilience/elastic.py
```diff
@@ -97,20 +125,25 @@ def resolve_resume_path(checkpoint_dir: str) -> Path | None:
...
+    for path in step_dirs:
+        if _dcp_durable(path):
+            logger.info(f"Auto-resume: found checkpoint at {path}")
+            return path
+        logger.warning(
+            f"Auto-resume: skipping {path} — no DCP .metadata, so the save that "
+            f"produced it did not complete"
+        )
 
     return None
```
amazloumi, line 148: Before this PR the function always returned a path when step_N directories existed. Now, if none of them looks durable, it returns nothing, and the entry path reads that as "no checkpoint to resume" and trains from scratch with no error.

kempnerforge/resilience/elastic.py
```diff
@@ -97,20 +125,25 @@ def resolve_resume_path(checkpoint_dir: str) -> Path | None:
...
         ),
         key=lambda d: int(d.name.split("_")[1]),
+        reverse=True,
     )
 
-    if step_dirs:
-        path = step_dirs[-1]
-        logger.info(f"Auto-resume: found checkpoint at {path}")
-        return path
+    for path in step_dirs:
```
amazloumi, line 139: Possible hang: Every rank now stats `.metadata`, which only the DCP coordinator writes, with no broadcast. The old name-based fallback always agreed across ranks. If one rank resolves None it returns early at `entry.py:207` and skips the collective `dcp.load` while the others block in it, hanging the job at startup.

kempnerforge/resilience/signal_handler.py
```diff
@@ -21,6 +21,33 @@
...
+
+    The cost: ``multiprocessing.Process.terminate()`` is itself SIGTERM, so a
+    shielded worker cannot be reaped that way and the interpreter's exit hook
+    would block joining it. Workers must therefore be shut down explicitly
+    through the loader's sentinel path — see ``BatchStream.close``.
+
+    Args:
+        worker_id: Worker index, supplied by ``DataLoader``. Unused.
+    """
+    for sig in _SHUTDOWN_SIGNALS:
```
amazloumi, line 47: in close(), w.kill() any worker still alive after _shutdown_workers().

PR #205, opened 2026-09-17
Reject sequence packing under pipeline parallelism

### LINKED ISSUE #204: Sequence packing is silently ignored under pipeline parallelism
## What happened

With `distributed.pp > 1` and `data.pack_sequences = true`, config validation accepts the job and training proceeds, but `doc_ids` never reaches the model — so packed documents attend across each other instead of being isolated.

Code trace (this was found by inspection, not from a failed run):

- `PipelineStageModule.forward` takes a single positional argument (`kempnerforge/distributed/pipeline_parallel.py:192`) and calls `layer(x, cos, sin)` at `:216` with no `doc_ids`.
- `pipeline_step` (`kempnerforge/training/loop.py:194-226`) reads only `batch["input_ids"]` and `batch["labels"]`. `batch["doc_ids"]` is present in the batch dict and dropped.

Compare the DP path at `kempnerforge/training/loop.py:308,317`, which does thread it into `model(input_ids, doc_ids=doc_ids)`.

## What I expected

Either block-diagonal attention isolating documents (as the DP path gives), or a validation error stating the combination is unsupported.

## Why it is easy to miss

Packing's *label* half still works under PP: labels are built in the dataset (`kempnerforge/data/dataset.py:62-64` sets `-100` at cross-document boundaries) and ride to the last stage as an ordinary tensor. The loss therefore correctly skips boundary predictions while attention leaks. Nothing crashes, and the loss curve looks plausible — if anything slightly better, since the model receives extra context it should not have.

## Reproduce

Before the fix, this is accepted rather than raising:

```python
JobConfig(
    data=DataConfig(pack_sequences=True),
    distributed=DistributedConfig(pp=2, dp_shard=1),
).validate(world_size=2)
```

## Severity

Latent rather than live: `pack_sequences` defaults to `false` and no config in `configs/train/` sets it. But the shipped PP configs (`13b_32gpu_tp4_pp2.toml`, `29b_32gpu_tp4_pp2.toml`, `70b_32gpu_tp4_pp4.toml`) would hit this the moment someone enables packing.

## Proposed fix

Reject the combination in `JobConfig.validate`. Actually threading `doc_ids` through the pipeline is a larger change — the stage module's forward signature plus per-microbatch alignment, since the schedule splits along dim 0 and does not pass a chunk id — and is worth tracking separately.

## Environment

torch 2.11.0+cu128, Python 3.13.5, KempnerForge `main` @ 58a2169.

timothyngo:
## Summary

Two independent, self-contained commits on the sequence-packing path. Neither depends on FlexAttention; this is the first of three PRs (see *Follow-ups*).

- **Reject `pp > 1` + `data.pack_sequences` in `JobConfig.validate`.** `PipelineStageModule.forward` receives only hidden states (`pipeline_parallel.py:192`, layer call at `:216`) and `pipeline_step` never read `batch["doc_ids"]` (`loop.py:194-226`), so the combination trained with cross-document attention. Packing's label half still worked — labels carry `-100` at boundaries from the dataset — so the loss looked correct while attention leaked, which is why it went unnoticed.
- **Vectorize `_compute_packed_output`'s doc-id assignment.** It carried a counter in a Python `for` loop over every token, inside `__getitem__` — `seq_len` interpreter iterations per sample, on the dataloader's critical path. `doc_ids[i]` is just the number of EOS tokens strictly before `i`, so one `np.cumsum` over the shifted EOS mask replaces it.

Kept as separate commits so the refactor is not bundled with the bug fix.

## Behavior change

The validation error is the only user-visible change. Packing without PP, and PP without packing, are both unchanged. No config in `configs/train/` sets `pack_sequences`, so no existing run is affected — the shipped PP configs (`13b_32gpu_tp4_pp2.toml`, `29b_32gpu_tp4_pp2.toml`, `70b_32gpu_tp4_pp4.toml`) all still load, which `test_shipped_config_loads` covers.

The vectorization is behavior-preserving: `TestComputePackedOutput` pins the semantics, and a new randomized test compares against the literal counter loop it replaces. I verified that test has teeth by mutating the implementation off-by-one — it fails, along with three of the existing cases.

## Testing

- [x] `uv run pytest tests/unit/` passes — **1808 passed, 13 skipped**
- [x] pre-commit (`ruff`, `ruff-format`) passes on all changed files
- [ ] `uv run pyright kempnerforge/` — 1 error, pre-existing and unrelated: `video_io.py:94` `import "av" could not be resolved`. `av` lives in the optional `video` dependency group, which `uv sync` does not install by default. **Zero errors in the files this PR touches.**

Not run: integration/distributed/e2e suites — this PR changes config validation and a numpy expression, touching no GPU or distributed code path.

One environment note for anyone reproducing on the cluster: `pytest --timeout=60` from CONTRIBUTING.md gives spurious failures in `tests/unit/test_packing.py` on a cold NFS cache, because `import transformers` walks its whole `models/` tree and can exceed 60s on its own. Those tests pass with a larger timeout; nothing to do with this change.

## Follow-ups

- **PR 2** — FlexAttention backend for packed sequences (`model.attention_backend = "flex"`), replacing the dense `(B, 1, S, S)` mask at `attention.py:214-227` that currently drops SDPA off FlashAttention whenever packing is on.
- **PR 3** — GPU validation matrix and throughput benchmarks for that path.
- Actually supporting packing under PP is separate work: the stage module's forward signature plus per-microbatch alignment, since the schedule splits along dim 0 and does not pass a chunk id. This PR's rejection is the assertion that change would invert.

Closes #204


timothyngo:
Verified the bug on hardware rather than leaving it as a code-reading argument: 2×H200, `pp=2` with `ScheduleGPipe`, fp32, batch 4 × seq 32 split into two 16-token documents.

Both paths run byte-identical weights — one `Transformer` built on CPU under a fixed seed, with the matching subset copied into each `PipelineStageModule` (stage submodule names and layer keys already match the full model). The script asserts the two key sets are exactly equal, so a silent partial load can't masquerade as a result. `JobConfig.validate` is bypassed deliberately, since the guard in this PR is the thing under test.

```
  PP vs single-GPU UNPACKED (control)   4.172e-07
  PP vs single-GPU PACKED               9.246e-01
  single PACKED vs UNPACKED  (sanity)   9.246e-01
```

Read together:

- **The control is clean** (4.2e-07, the fp32 noise floor), so the PP machinery, the weight copy, and microbatching are all correct. Whatever the second line shows is not an artifact of the harness.
- **PP's output *is* the unpacked output.** Not approximately: `PP vs PACKED` and `PACKED vs UNPACKED` are the same number to four significant figures. PP lands exactly on plain causal attention.

So `doc_ids` is not partially mishandled under pipeline parallelism — it is dropped entirely, and every document in a packed window attends to every earlier one. `PipelineStageModule.forward` takes only hidden states, and `pipeline_step` never reads `batch["doc_ids"]`.

The reason this is worth a hard `raise` rather than a warning: packing's *label* half keeps working under PP, because labels carry `-100` at document boundaries from the dataset and travel to the last stage as an ordinary tensor. The loss therefore looks correct — arguably *better*, since the model is getting context it should not have — while attention leaks across every boundary. There is no signal in the training curve that would reveal it.

Nothing shipped is affected: `pack_sequences` defaults to `false` and no config in `configs/train/` enables it. The exposure is the next person to turn packing on for one of the existing PP recipes (`13b_32gpu_tp4_pp2.toml`, `29b_32gpu_tp4_pp2.toml`, `70b_32gpu_tp4_pp4.toml`).

Follow-up, not in this PR: actually threading `doc_ids` through the pipeline. That needs a second argument on the stage-module forward plus per-microbatch alignment, since the schedule splits along dim 0 and does not pass a chunk id. The rejection added here is the assertion that change would invert. The *control* line above is also a regression test worth keeping on its own — there is no PP-vs-single-GPU equivalence test in `tests/distributed/` today — so it lands in the validation PR later in the stack.

kempnerforge/config/job.py
```diff
@@ -197,6 +197,19 @@ def validate(self, world_size: int = 1) -> None:
...
             )
 
+        # Why this is an error rather than a warning: PipelineStageModule.forward
+        # receives only hidden states, so doc_ids never reaches the stages and
+        # packed documents attend across each other. The labels still carry -100
+        # at the boundaries, so the loss looks correct while attention leaks --
+        # there is no signal in the training curve that would reveal it.
+        # Measured on 2 GPUs: pp=2 output matches unpacked causal attention
+        # exactly, rather than the packed reference.
+        if self.distributed.pp > 1 and self.data.pack_sequences:
```
amazloumi: The message currently only says `not supported`. Please move the reasoning into the error message and remove the comment. Also remove `Measured on 2 GPUs`; experiment details shouldn’t live in source code.

More generally, We want minimize comments and keep only docstrings in the source. We have one todo task for this cleanup.

amazloumi: Could this move to `JobConfig.__post_init__`? as stated in CONTRIBUTING.md we prefer guards to fail fast while currently costs a full process-group init before it fires.

tests/unit/test_packing.py
```diff
@@ -89,6 +89,27 @@ def test_output_dtypes(self):
...
+            return doc_ids
+
+        rng = np.random.default_rng(0)
+        for _ in range(50):
+            # Small vocab so EOS is dense: boundaries, runs of consecutive EOS,
+            # and EOS at either end all show up across the draws.
+            tokens = rng.integers(0, 4, size=int(rng.integers(3, 64))).astype(np.int64)
+            result = _compute_packed_output(tokens, eos_token_id=0)
+            # _compute_packed_output returns the input half of the doc ids.
+            assert result["doc_ids"].tolist() == reference(tokens, 0)[:-1].tolist()
```
amazloumi: Could this assert on labels too? As written it doesn't catch a tail off-by-one in the line it guards: `result["doc_ids"]` is `doc_id_tensor[:-1]`, so the last doc id is never inspected, and that element is what masks the final label.

could be something like:

```
              ref = reference(tokens, 0)
              ...
              expected = tokens[1:].copy()
              expected[ref[:-1] != ref[1:]] = -100
              assert result["labels"].tolist() == expected.tolist()
```

kempnerforge/data/dataset.py
```diff
@@ -41,13 +41,12 @@ def _compute_packed_output(tokens: np.ndarray, eos_token_id: int) -> dict[str, t
...
+    # doc_ids[i] is the number of EOS tokens strictly before i. Vectorized
+    # because this runs per __getitem__, once per token -- a Python loop here
+    # costs seq_len interpreter iterations on every sample the loader yields.
     doc_ids = np.zeros(len(tokens), dtype=np.int64)
-    doc_id = 0
-    for i in range(len(tokens)):
-        doc_ids[i] = doc_id
-        if tokens[i] == eos_token_id:
-            doc_id += 1
+    doc_ids[1:] = np.cumsum(tokens[:-1] == eos_token_id)
```
amazloumi, line 49: Minor: this now requires an ndarray, where the loop worked for any sequence.

amazloumi (commented):
Good find, and the diagnosis is right. Confirmed on 2 H200s: main accepts pp=2 + packing and trains six steps with no error, this PR raises instead, and pp=2 with packing off still trains, so the check doesn't over-reach. The vectorization is integer-identical to the loop and the packing/config tests pass.

Just three inline comments: the check belongs in `__post_init__` per CONTRIBUTING.md, the error message should carry the reasoning, and the new test doesn't catch a tail off-by-one. One note only: this raise also fires in `scripts/eval.py`, which is non-PP and never packs.

kempnerforge/config/job.py
timothyngo: Has this been the standard for this repo? It's pretty typical to have short error messages, since end-users don't need to know the reason for why something is not supported. The reason should be a comment.

I think we should add comments to the source. The comments are useful for capturing details for models and humans without polluting docstrings with details that should not be in public documentation.

amazloumi: I see your point. I’m not suggesting that we eliminate comments entirely. They should be brief and included only when necessary. This improves maintainability and avoids adding information that can quickly become outdated, potentially misleading agents that read the code. Ideally, the source code should remain the agents’ primary, and preferably sole, source of truth.

kempnerforge/data/dataset.py
timothyngo: The original code already required `tokens` to be an ndarray (see line 52 in the original code).

kempnerforge/config/job.py
timothyngo: I can clean up the comment and link to the corresponding issue, but I think the exception should stay as is.

timothyngo: Yes, can do that.

timothyngo: Done

tests/unit/test_packing.py
timothyngo: Done

kempnerforge/config/job.py
timothyngo: Done

amazloumi (approved):
Thanks. LGTM

PR #207, opened 2026-09-20
Validate and benchmark the FlexAttention packed path

timothyngo:
## Summary

Tests, benchmarks, and docs for `attention_backend="flex"`. **No behaviour change** — no file under `kempnerforge/` is touched.

> **Stacked on #206**, which is stacked on #205. Review order is 205 → 206 → this.

## The benchmark

One H200, 125M params (`dim=768`, 12 layers), forward+backward, 8 documents per sequence:

| seq_len | `sdpa` packed | `flex` packed | speedup | peak memory |
|---|---|---|---|---|
| 512 | 278k tok/s | 199k tok/s | 0.72x | 4.97 → 4.87 GB |
| 2048 | 236k tok/s | 310k tok/s | **1.31x** | 10.16 → 9.36 GB |
| 8192 | 124k tok/s | 290k tok/s | **2.34x** | 12.60 → **9.38 GB** |

The row that reframes the whole feature is `seq_len=8192`: the dense-mask path runs at **124k tok/s against 236k unpacked**. Packing has been *costing* throughput, not saving it — enabling it made runs slower. Flex reaches 290k, which is faster than not packing at all, because block-diagonal attention does strictly less work than full causal. So #206 is less an optimization than the thing that makes an existing feature worth switching on.

Flex *losing* at 512 (0.72x) is reported rather than hidden: below roughly a thousand tokens there is little block sparsity to exploit and the mask construction does not pay for itself. That is why the `seq_len >= 128` floor in #206 is a hard floor and the docs steer toward long sequences.

Two secondary sweeps: more documents makes the mask sparser, so flex improves while the dense path does not move at all (296k / 313k / 313k for 2 / 16 / 64 documents at `S=2048`, against a flat 236k); and GQA widens the gap to 1.37x, since flex passes `enable_gqa` rather than materializing repeated K/V heads.

## The compositions #206 argued for but never ran

#206 reasoned about these from the code and shipped without exercising them. Now they are checked:

- **`tests/integration/test_flex_ac.py`** — `{none, selective, full}` activation checkpointing × `{eager, compiled}`, all six. This nests flex's higher-order op inside the checkpoint higher-order op inside the model graph. Forward parity, gradient parity, and document isolation *across the recompute*. The SDPA reference is deliberately built without AC: checkpointing is a recompute strategy that must not change the answer, so folding it into the reference too would mask the failure worth catching.
- **`tests/distributed/test_tp.py::TestTensorParallelFlexAttention`** — parity against the dense-mask path under TP, cross-rank agreement after the row-parallel all-reduce, document isolation with sharded heads, and backward. #206 argued correctness under TP from `H=None` making the BlockMask head-broadcast and `n_rep` coming from global head counts; this makes that checkable rather than merely plausible.
- **`tests/distributed/test_pp.py`** (new) — pipeline-parallel output vs a single-GPU forward over identical weights (`pp=2`, `ScheduleGPipe`). The repo had no PP equivalence test of any kind. It is deliberately *unpacked*: `doc_ids` never reaches pipeline stages, which #205 now rejects. This is the control from the investigation that established the leak — it passing at 4.2e-07 is what proves the PP harness itself is sound, and therefore that the packed mismatch reported on #205 is real rather than an artifact.

## Docs

`docs/architecture/model.md` gains an *Attention backend* section carrying the measured table, and grows from three attention paths to four. `docs/how-to/prepare-tokenized-data.md` gains backend guidance plus two caveats worth knowing before enabling packing: PP is unsupported, and **eval never packs** — `_EvalTensorDataset` emits no `doc_ids`, so eval loss is computed under different masking than training loss. That second one is pre-existing and previously undocumented.

## Testing

- [x] `uv run pytest tests/unit/` — **1843 passed**, 13 skipped
- [x] `pytest tests/integration/test_flex_ac.py` on 1×H200 — **11 passed**
- [x] `torchrun --nproc_per_node=2 -m pytest tests/distributed/test_pp.py tests/distributed/test_tp.py` on 2×H200 — **8 passed**, 1 skipped (TP+FSDP needs ≥ 4 GPUs)
- [x] `ruff check` / `ruff format --check` / pre-commit pass
- [x] `pyright kempnerforge/` — only the pre-existing `video_io.py` `import av` error (optional `video` group, not installed)

One note on scope hygiene: `ruff format benchmarks/` also reformats `bench_moe.py`, which is unrelated to this work and drifted because `benchmarks/` is outside CI's lint scope (`kempnerforge/ tests/ scripts/`). That reformat is deliberately **not** included here — worth its own commit if anyone wants it.

## Follow-ups

- Flip the default to `"flex"` for packed runs, now that there are numbers to justify it.
- Thread `doc_ids` through pipeline parallelism and invert #205's rejection.
- Per-document RoPE position reset, which is a semantics change and belongs on its own.

Refs #204


timothyngo:
Added FSDP2 coverage (`4f34413`), closing the last "argued but not run" gap from #206.

The interaction worth checking is `default_mp_policy`'s `cast_forward_inputs=True`. FSDP2 walks forward inputs at each wrapped-module boundary and casts them to bf16 — but `doc_ids` is an int64 *index* tensor, and a `BlockMask` is not a tensor at all. If either got mangled on the way in, nothing would raise: the mask would quietly widen and documents would stop being isolated, producing perfectly plausible logits. Same failure mode as the `torch.compile` bug fixed in the parent PR, which is why these assert **exact equality** rather than a tolerance — document 1 may not move *at all* when document 0 changes.

| test | covers |
|---|---|
| `test_flex_matches_sdpa_under_fsdp` | parity vs the dense-mask path (bf16 tolerance) |
| `test_doc_ids_survive_the_forward_input_cast` | int64 `doc_ids` through the bf16 cast — exact |
| `test_explicit_block_mask_survives_fsdp_input_cast` | the `block_mask=` argument, the one route that hands a `BlockMask` to a sharded module — exact |
| `test_flex_backward_under_fsdp` | gradients finite under sharding |
| `test_flex_with_activation_checkpointing_under_fsdp[selective, full]` | the full AC + FSDP2 + flex production stack |

**6 passed on 4×H200**, 17 total in `test_fsdp.py`. Each was confirmed individually rather than inferred from the aggregate count — the first run's log was truncated and only showed three of them, so I re-ran the class with untruncated output before claiming it.

That leaves one gap still open on this stack, stated plainly: no end-to-end training run. The correctness case rests on kernel-, model-, and now parallelism-level parity, not on a loss curve.

amazloumi (approved):
Approving. Worth re-running the sweep once the earlier PRs in the stack land, in case the table needs an update before merging.

